PR #17, opened 2014-12-31
Add tutorial subsection about type signatures for `Default` instances

c-lewis:
The need for type signatures needs to be called out. I sometimes forget this---especially when using `leftJoin`.


tomjaguarpaw:
Thanks, that's helpful.  I edited the explanation and put it with the `leftJoin` example rather than in the introduction, and I also made the issue explicit in the Haddocks.  Let me know what you think.

https://github.com/tomjaguarpaw/haskell-opaleye/commit/10d1a7725d7855c6b649a218c2c4c29304808504

c-lewis:
Looks good to me. Thanks for including this.

PR #18, opened 2015-01-08
Remove extra output value in SQL for join

johntyree:
The idealized SQL for the join example is wrong as the query should be returning four values. Just removing the extras here.


tomjaguarpaw:
That's right, thanks.

PR #25, opened 2015-03-03
Experimental branch with Postgres removed

rehno-lindeque:
I thought I'd open this PR mostly for discussion. As you know I'm looking at using Opaleye in conjunction with Persistent (I'm thinking, as a replacement for esqueleto initially), using [rawQuery](http://hackage.haskell.org/package/persistent-2.1.1.7/docs/Database-Persist-Sql.html#v:rawQuery), [rawExecute](http://hackage.haskell.org/package/persistent-2.1.1.7/docs/Database-Persist-Sql.html#v:rawExecute) and [rawSql](http://hackage.haskell.org/package/persistent-2.1.1.7/docs/Database-Persist-Sql.html#v:rawSql) for now.

I wonder what your plans are for if a mysql backend were added? Would this split into `opaleye`,`opaleye-postgresql`,`opaleye-mysql` packages?


bergmark:
I've taken part in discussions about other backends before. I only use postgresql and I'm hesitant about supporting other backends. It might mean complicating the code base, and in the worst case it would mean that the features of opaleye would be limited to the intersection of features supported by the various backends.

My biggest fear would be that a postgres feature couldn't be added because mysql doesn't support it. I'm also not interested in making sure any changes I make also work for mysql.

If these turn out to not be issues, then I wouldn't mind!

tomjaguarpaw:
Hi @rehno-lindeque, thanks for the PR.  I will look at it when I get a free moment.

To quickly clarify my policy:

Opaleye is a pure Postgres library and support for Postgres will never be compromised for the purpose of supporting other backends.  I am very interested in starting *similar* projects for MySQL, SQL Server and SQLite, but they will be different libraries.  Implementation details may be shared to the extent that it does not compromise supporting a full set of features for each RDBMS.

rehno-lindeque:
Thanks @bergmark and @tomjaguarpaw both, that sounds reasonable to me. I'd love to migrate to Postgres myself and have the best possible feature set available. I'll close this PR; it only removes `postgresql-simple` as a dependency (a quick experiment on my part, trying to get a vague idea for how far off a MySQL port would be).

There were only two superficial issues that jumped out in the attempt:
* The first issue I picked up was simply to do with quoting identifiers. Postgres uses double quotes `"field"` whereas MySQL uses backticks \``field`\`
* The second was a generic query optimization that I suppose could be helpful for debugging in general https://github.com/tomjaguarpaw/haskell-opaleye/issues/26

I couldn't say what else might crop up, it seems as though there should be a fairly significant amount of code that could be shared anyway. Cheers!

tomjaguarpaw:
Changing the quoting method will be very easy.  Query optimization is a large problem in general, but hopfully an easy improvement can be found that works in the vast majority of cases.

PR #32, opened 2015-04-29
Add json types

c-lewis:
The addition of these types intentionally does not address a number of issues:

* A mechanism, if there should be any, for indicating that these types are not available in all currently supported versions of PostgreSQL (currently, comments in the code indicate the applicable versions). A mechanism for indicating the versions of PostgreSQL to which various types, etc. apply is outside the scope of this change.

* The streaming of large documents to the database. It's not clear to me how this would work or should be done.

* Automatic decoding of types in a JSON serialization type class by the `pg*JSON[B]` column generation functions. This would require a dependency on a JSON serialization package (`aeson`, `json`, etc.), and I for one prefer to keep package dependencies to a minimum, especially in situations like this, where only a small subset of the depending package's consumers will utilize the functionality requiring the dependency.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -64,6 +64,7 @@ library
                    Opaleye.Internal.Order,
                    Opaleye.Internal.Optimize,
                    Opaleye.Internal.PackMap,
+                   Opaleye.Internal.PGTypes,
                    Opaleye.Internal.PrimQuery,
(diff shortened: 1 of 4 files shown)


tomjaguarpaw:
Hi Christopher, thanks for that.  It looks good.  Let me find some time to look it over properly and then I'll merge it.

Issue #14, opened 2014-12-22
Aggregate min on Bool column generates invalid query


coubeatczech:
Running `Aggregate.min` function on boolean column generates an invalid query that will fail when ran.

    select id, min(closed) from
    No function matches the given name and argument types. You might need to add explicit type casts.

Some other functions from the same module might have the same problem, like `avg` for string fields etc.

------------------------

**Idea:** It could be solved by restricting the `a` type of `min` with a methodless typeclass. All the datatypes that can be applied to the function would be instances of the typeclass.

tomjaguarpaw:
https://github.com/tomjaguarpaw/haskell-opaleye/commit/be092f3948c56be419f1fec93f2554e5049c0d8c

PR #33, opened 2015-05-05
I14

coubeatczech:
Implementation of #14 

I actually tried in the `psql` 9.3.6, if the aggregation doesn't result exception for the types that have the `ColumnComparable` typeclass.

I'd like to cover the other types if this is ok, the single-dimensional array of ints works, there might be rule that if ColumnComparable a => ColumnComparable (DBArray a), but that would need some testing.

### DIFF
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -39,11 +39,11 @@ avg :: Aggregator (C.Column T.PGFloat8) (C.Column T.PGFloat8)
 avg = A.makeAggr HPQ.AggrAvg
 
 -- | Maximum of a group
-max :: Aggregator (C.Column a) (C.Column a)
+max :: (T.ComparableColumn a) => Aggregator (C.Column a) (C.Column a)
 max = A.makeAggr HPQ.AggrMax
 
 -- | Maximum of a group
-min :: Aggregator (C.Column a) (C.Column a)
+min :: (T.ComparableColumn a) => Aggregator (C.Column a) (C.Column a)
 min = A.makeAggr HPQ.AggrMin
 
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
Thanks for that.  I will give it a look when I have some free time.

bergmark:
Can this class just be used for aggregations or is it the same set of types as for normal comparisons (`>` etc)?

coubeatczech:
Good point! I guess it should work for the normal comparisons too, the typeclass would need to be changed though as the normal comparisons like `.==` takes two types of columns, therefore having only one parameter would forbid the user to compare say `PGInt4` and `PGFloat8` without some kind of casting.

coubeatczech:
Currently, it is only allowed to compare columns of the same type, so I added an optional commit with type restriction in comparison: ea51e26

bergmark:
Allowing to compare columns of different type would force you to use monomorphic functions to apply constants right? `x .== pgInt4 1` would be okay but `x .== constant 2` ambiguous. I'd rather have a concept of safe "casts" for this: `(x :: Column PGInt8)  .== safeCast (y :: PGInt 4)`. Cast is probably not a good name since there is no explicit casting needed in the generated SQL...

coubeatczech:
Well that's what one need to do now, to call the `pgXXX` when comparing to constant to a column, no? I guess that's another issue. I think this issue doesn't touch it, it would be good to avoid the `pgXXX` explicit calls though.

tomjaguarpaw:
To clarify, `(.==)` will never allow specifying arguments of different types, and neither will the order comparison operators.

tomjaguarpaw:
https://github.com/tomjaguarpaw/haskell-opaleye/commit/be092f3948c56be419f1fec93f2554e5049c0d8c

coubeatczech:
Cool!

PR #48, opened 2015-06-19
Wrap table names in double quotes

tstat:
This adds support for tables whose names include capital letters.


tomjaguarpaw:
Thanks, that looks good.

It can't be automatically merged, presumably because I just merged your whitespace patch.  Can you amend and resubmit?

tstat:
I resolved the conflict and force pushed over my branch. Everything should be up to date now.

tomjaguarpaw:
Thanks!

PR #49, opened 2015-06-19
Clean white space related issues.

tstat:
I was reading through some of Opaleye and noticed there was often mixes of between the use of hard tabs and spaces within the same declaration or line. There was also a lot of trailing white space. This commit just eliminates all trailing white space and replaces tabs with spaces.


tomjaguarpaw:
Thanks that's very helpful.  It was mostly stuff I copied from HaskellDB and then never bothered to change.

PR #53, opened 2015-06-21
I52

coubeatczech:
This adds the **array_agg(expression)** operator + a simple test.


tomjaguarpaw:
That's great, thanks!

PR #57, opened 2015-06-30
Fix warnings.

hesselink:
This fixes almost all warnings. The only one remaining is about 'join'
clashing with a future prelude, but that one requires some decision by
you, since renaming it will break user code, but leaving it in
probably also will.


tomjaguarpaw:
`join` is not publically visible so you can just go ahead and change it to `join_` if you like.

PR #58, opened 2015-06-30
Adding support for WITH RECURSIVE.

hesselink:
This isn't done, but I'd like some feedback. I've got most of the
plumbing in place, but there are some issues. One is that I need to
add the table alias introduced by WITH RECURSIVE to the second
(recursive) case. I've done this now in
Opaleye.Internal.Sql.withRecursive, but I feel it's not ideal. It also
doesn't quite work, since it generates something like this:

``` sql
SELECT "id0_2" as result1_4,
       "name1_2" as result2_4,
       "subpart2_2" as result3_4
FROM (SELECT *
      FROM (WITH RECURSIVE "with_recursive_result_3" AS
            (SELECT *
             FROM (SELECT "id" as id0_1,
                          "name" as name1_1,
                          "subpart" as subpart2_1
                   FROM "parts" as T1) as T1
             UNION ALL
             SELECT *
             FROM "with_recursive_result_3" as T1,
                  (SELECT "id" as id0_2,
                          "name" as name1_2,
                          "subpart" as subpart2_2
                   FROM "parts" as T1) as T2
             WHERE (("id0_1") = ("subpart2_2")))
            SELECT * FROM "with_recursive_result_3") as T1) as T1
```

Notice that the second, recursive query adds the recursive table, but
does a 'SELECT _' outside of that, causing it to select more columns
than is wanted. I could either try to not select `_` but the right
columns, or try to push the recursive table a bit further in (but how
far?).

Any advice, pointers and answers to my questions appreciated :)


hesselink:
Another thing I realized I have to fix: I'm not naming the columns in the `WITH RECURSIVE` statement, which I should do, and then the outermost select should use those column names.

coubeatczech:
In order to select just the correct columns from the recursive - maybe you can extract the `PrimExpr`s from the columns that are returned from the recursive query and attach that info to the `WithRecursive` data constructor of `PrimQuery` and then wrap the `WithRecursive` constructor in a `Select` that will select only the attributes attached to the `WithRecursive`.

tomjaguarpaw:
If you want to know which column names should appear in the select I think you will want to use `Unpackspec`.

hesselink:
@coubeatczech Yes, that's the direction I'm currently working in, but I was having trouble finding the column names. Perhaps @tomjaguarpaw's `Unpackspec` suggestion is the thing I was missing, I'll have a better look when I get time to work on this again (perhaps tonight).

tomjaguarpaw:
`Unpackspec` is not documented, but see `Opaleye.Internal.QueryArr.runQueryArrUnpack` for how the `Unpackspec` is used to unpack the column names.

tomjaguarpaw:
See also https://github.com/tomjaguarpaw/haskell-opaleye/issues/61

hesselink:
OK, that looks great. The `PrimExpr` I get back can be anything, right? So I'll have to write a traversal of that to get the column names.

tomjaguarpaw:
Yes the `PrimExpr` is expressions with column names at the leaves.  I think I must be not understanding exactly what you want and will have to take a closer look.

hesselink:
Oh, it's guaranteed to have column names at the leaves? That's exactly what I want.

tomjaguarpaw:
Without having looked to hard at what you're trying to do, I would guess you want to take the `PrimExpr`s as they are and come up with fresh column names for them.

hesselink:
OK, that sounds reasonable. So in the places where I now have a `*`, I could generate something like `SELECT old1 as fresh1, old2 as fresh2, old3 as fresh3`.

coubeatczech:
@hesselink look at this: https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/PackMap.hs#L72

I think this is the traversal you want to use in order to create fresh names.

tomjaguarpaw:
@hesselink Yes, though probably more like `old1 + old2 * old3 as fresh1`, but `old1 + old2 * old3` is just the `PrimExpr` so you don't have to do anything with it, you just copy it over.

tomjaguarpaw:
I documented `Unpackspec` a bit.

tomjaguarpaw:
And as @coubeatczech says, if you want a neat way of making fresh names look at the `PM` monad in the `PackMap` module.

hesselink:
OK, I finally had some time to work on this again, and I have a working version! I'd appreciate any comments on the code and implementation. A couple of small questions:

* I do a `runUnpackspec` twice in `withRecursive`: once to get the column names from the recursive query, and again to get the (fresh) column names to use for the total recursive query. Is there a simpler way?
* I pass in `newCols`, the final result, as the argument to running the recursive query arrow. This makes sense since they construct is meant to be recursive, but I don't know enough about opaleye to know if this is safe in general, or if it might loop in some cases.
* The `addTable` function is the messiest part, I think, but I don't see any other way to do this.

tomjaguarpaw:
Cool!  It'll be a few  days before I can look at this properly.

> I do a runUnpackspec twice in withRecursive: once to get the column names from the recursive query, and again to get the (fresh) column names to use for the total recursive query. Is there a simpler way?

You can come up with an `Applicative` (`Sum f g a = Sum (f a) (g a)` would do) that does both things that you want in one go, but I probably wouldn't bother.

tomjaguarpaw:
I've had a quick look through the patch.  The style in very nice and in keeping with the rest of the codebase, so thanks for paying attention to that.

I will have to spend some more time looking at it as I am not familiar with `WITH RECURSIVE`.  There are now QuickCheck property tests, so you might want to add some for it.

Minor comment

* `f pe = ([pe], pe)` etc. is now defined here: https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/Unpackspec.hs#L50

hesselink:
Thanks! I've rebased the branch on the latest master, used `collectPEs`, exported `Opaleye.WithRecursive` from the `Opaleye` module, and implemented a simple test case.

Can you briefly explain what the QuickCheck tests test, exactly? Are they basically testing a Haskell implementation of the tested function against an Opaleye version?

tomjaguarpaw:
The QuickCheck tests are supposed to be testing the denotational semantics of Opaleye, those being that a `Query a` is supposed to correspond very closely to a `[a]`.  In practice though your description "testing a Haskell implementation of the tested function against an Opaleye version"  is essentially the same thing.

tomjaguarpaw:
I'm wondering whether the type `QueryArr a a -> QueryArr a a` makes sense, as well as `(Query a -> Query a) -> (Query a -> Query a)`.  I'll have to give it more thought.

tomjaguarpaw:
Have you been using this at all?  It will be some time before I can give it a thorough look I'm afraid.

hesselink:
I've generated one query we would use in our app and verified that it does what I expect, but we're not running it live (yet). I feel pretty good about the patch in general, but there is one part (the `addTable` function) where I still have to at least run all of the branches once. I think implementing a quickcheck test will do this, so that's what I'm working on now (when I find the time).

rvl:
I need a simple recursive query in my app. This branch seems to work in my case, after merging with latest master (see 338ca19). It would be really nice to have basic recursive queries in the next version.

rvl:
I was wondering if the parameters to `withRecursive` should be flipped? For example, in my app I do the same recursive query, but the base query is quite variable, e.g. joins on different things.

hesselink:
@rvl I'll try to get the branch merged soonish. @tomjaguarpaw Do you think you'd accept this patch without further work? I was planning to add more tests, but I don't think I'll have time to do it. But it works as far as I've tested it, and apparently it also works for @rvl.

@rvl Re: the parameter order: I only have one use case, which didn't partially apply it, so I just chose a parameter order that matches SQL. @tomjaguarpaw What do you think?

tomjaguarpaw:
@hesselink Are you running this live?  I'm happy to add it but I want to make 100% the type signature makes sense.

hesselink:
No, I'm not running it live. I built it for something where it worked, but it turned out we didn't need the feature. But back then it did work :) And it does have one test case.

I'm not sure how to know if the type signature makes sense. I'm not sure if it's more common to vary the base case or the recursive case. One point for the current order is that the base case is usually much shorter, so you might be able to write `withRecursive base $ something very long for recursive`. But the only real way to find out is by using it, and it can't be used without being released ...

tomjaguarpaw:
Oh, I meant "makes sense" in the sense that it doesn't leak the abstraction.

hesselink:
Oh ok, that I'm not sure about. Implementing it, I had to use recursion in exactly where you'd expect it, which was the recursive query. I'll leave you to think about it, and I'll rebase and remove the TODOs when I have the time.

rvl:
Possibly someone might want to use `UNION` instead of `UNION ALL`.

I also tried changing the signature to
```
withRecursive :: Default Unpackspec a a => QueryArr i a -> QueryArr a a -> QueryArr i a
```
so that joins with the base query could be possible. But PostgreSQL didn't like the resulting query. Actually I'm not even sure if there could always exist valid sql corresponding to the above signature.

If merging now, the `withRecursive` API could be marked as provisional and subject to change. Or on second thoughts it could be deferred until more users are found. Because using stack and/or Nix, it's fairly easy to run a forked version of a dependency.

hesselink:
I've merged the PR so this contains the latest master now. I've also removed the TODOs as I won't have time to add extensive tests any time soon. I do feel pretty ok about the code though, as I mentioned.

Regarding the type signature: I think the current one is at least correct. It takes a base query (`Query a`) and a recursive query with takes the result of the base query as input (`QueryArr a a`). This is exactly the semantics and syntax of `WITH RECURSIVE` in postgresql. There maybe be more relaxed type signatures, but I'm not sure about those, so this signature seems to be a safe start.

@tomjaguarpaw: If you want to merge this but not advertise it too much, perhaps we could remove the re-export in `Opaleye`?

tomjaguarpaw:
@rvl When you merged it looks like you added `testKeywordColNames` and `testInsertSerial` to the test list a second time.

tomjaguarpaw:
That's why the tests are failing.

rvl:
OK sorry about that, I was actually testing with most tests commented out.

I've come across a `withRecursive` query which PostgreSQL doesn't like: "each UNION query must have the same number of columns".

I'm not sure how to address it. What do the type class constraints on the `unionAll` function do? Would that be related to this?

```haskell
testWithRecursive2 :: Test
testWithRecursive2 = testG q (== expected)
  where
    q = O.withRecursive baseQ recQ
    -- Join table3 and table8
    baseQ = proc () -> do
      r3 <- O.queryTable table3 -< ()
      r8 <- O.queryTable table8 -< ()
      O.restrict -< fst r8 .== snd r3
      Arr.returnA -< r8
    -- And everything where 'column2' equals a selected 'column1',
    -- recursively.
    recQ = proc base -> do
      recursive <- O.queryTable table8 -< ()
      O.restrict -< O.toNullable (fst base) .== snd recursive
      Arr.returnA -< recursive
    expected :: [(Int, Maybe Int)]
    expected = [ (1, Nothing)
               , (2, Just 1)
               , (4, Just 2) ]
```

tomjaguarpaw:
Well, they make sure they both have the same columns, so yes I think it would be very related :)

I don't have time to look at this now and admittedly the `BinarySpec` code is very arcane but perhaps you can figure it out.

tomjaguarpaw:
Superceded by https://github.com/tomjaguarpaw/haskell-opaleye/commit/6cc38b455e1250d33e2a1e1b85086710a0c25cee

Issue #52, opened 2015-06-21
Expose/implement more aggregation operators


coubeatczech:
In the postgresql, there are more aggregation operators than the user of opaleye can use. Some are just not exposed in the public api, whereas some are just not implemented.

PR #59, opened 2015-06-30
parametrized aggregations #52

coubeatczech:
Add support for aggregations that take more arguments than just the
argument to aggregate
- implement the string_aggr aggregation
- test the string_aggr

Also, having another argument in the `AggrExpr` constructor in the `PrimExpr` type could be expressed as adding a `Maybe PrimExpr` or making it `NonEmpty` list.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -329,10 +329,17 @@ testAggregateProfunctor = testG q expected
                            (\(x, y) -> aggregateCoerceFIXME' x * y)
                            (PP.p2 (O.sum, O.count))
 
-testStringAggregate :: Test
-testStringAggregate = testG q expected
+testStringArrayAggregate :: Test
+testStringArrayAggregate = testG q expected
   where q = O.aggregate (PP.p2 (O.array, O.min)) table6Q
         expected r = [(map fst table6data, minimum (map snd table6data))] == r
+
+testStringAggregate :: Test
+testStringAggregate = testG q expected
+  where q = O.aggregate (PP.p2 ((O.string . O.pgString) "_", O.groupBy)) table6Q
+        expected r = [(
+          (foldl1 (\x y -> x ++ "_" ++ y) . map fst) table6data ,
+          head (map snd table6data))] == r
 
 testOrderByG :: O.Order (Column O.PGInt4, Column O.PGInt4)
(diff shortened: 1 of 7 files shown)


tomjaguarpaw:
I would have thought that the delimeter should be an argument to the `AggrStringAggr` constructor.

Do you know what the delimiter parameter is allowed to depend on?  Is it allowed to depend on some columns from the query, or does it have to be a constant expression?

coubeatczech:
It cannot use be a result of aggregation function ran on the columns, so it will typically be just a `ConstExpr`.

So you suggest having `AggrStringAggr PrimExpr` constructor instead?

tomjaguarpaw:
> So you suggest having `AggrStringAggr PrimExpr` constructor instead?

Yes, that seems like the right way to do it.  It will require rearranging some of the internals though.

tomjaguarpaw:
Can you please check precisely the scope allowed for the delimiter?  At the moment it will permit

    proc () -> do
        delim <- table1 -< ()
        aggregate (string delim) table2 -< ()

so we have to be sure that is allowed.

coubeatczech:
Well you can't have the code you have shown in the snippet, no? The `delim` is not in the scope, it's not translateable into `>>>`'s and `first`'s.

tomjaguarpaw:
Err, yes you're right!  So then the question becomes the converse: if it is allowed we need to think of a way we can support it.  (However that question is not nearly as pressing).

coubeatczech:
If this was allowed, then a query similar to this one would be generated:

    select * 
    from 
      (select * from t1) as t , 
      (select (string_agg(value, t.value)) from t1 group by id) as tt

But, a query like this can't be run in psql:

    ERROR:  invalid reference to FROM-clause entry for table "t"
    LINE 4:   (select (string_agg(value, t.value)) from t1 group by id) ...
                                         ^
    HINT:  There is an entry for table "t", but it cannot be referenced from this part of the query.

So I think that it's matching the functionality that the psql provides.

tomjaguarpaw:
Great, thanks for doing that check.  So it looks like your form is the most general permissible.  This is another win for Opaleye's type safety!  I'll merge this tomorrow, hopefully.

Issue #21, opened 2015-02-16
Lacking `in` operator


ocharles:
I had a stab at this with:

```haskell
in_ :: Column a -> [Column a] -> Column PGBool
in_ x ys =
  Opaleye.binOp HPQ.OpIn
                x
                (Opaleye.Column (HPQ.ListExpr [y | Opaleye.Column y <- ys]))
```

However this produces:

```sql
WHERE ((id0_2) IN ((10025,10355)))) as T1
```

Which isn't correct, as there are one too many levels of parenthesis, which changes the semantics to be a list of records, rather than list of integers.

bergmark:
This is what we've been using, is it good enough?

```haskell
ors :: Foldable f => f (Column PGBool) -> Column PGBool
ors = foldr (.||) (constant False)

ands :: Foldable f => f (Column PGBool) -> Column PGBool
ands = foldr (.&&) (constant True)

inE :: [Column o] -> Column o -> Column PGBool
inE hs w = ors . map (w .==) $ hs

notIn :: [Column a] -> Column a -> Column PGBool
notIn hs w = ands . map (w ./=) $ hs
```

ocharles:
It's definitely good enough, but it might be nice to have something that maps directly to the `or` operator. If we don't do that, we should get @bergmark's functions added.

tomjaguarpaw:
@ocharles What do you mean "maps directly to the `or` operator"?

bergmark:
I'm guessing he meant `s/or/in`...

ocharles:
Bleh, I did indeed. I would expect the compiled query to be:

```sql
x IN (a, b, c)
```

or

```sql
x IN null
```

tomjaguarpaw:
Is there a particular benefit to that beyond the stylistic?

ocharles:
I don't have a good argument for that, so if you think folding `OR` is better than we can do that. My main desire is that this exists in `opaleye` itself. I probably lean towards `IN` because that's being even more declarative to the query planner about what you want, which may give it an easier job finding an efficient plan. I have nothing to back that up, however.

tomjaguarpaw:
Folding `OR` is easier to implement, since, as you have seen, parenthesising an SQL expression can cause a change in meaning.  As you say, `IN` might lead to better performance.

[...]

PR #66, opened 2015-07-05
implement IN and NOT IN the terms of OR and AND #21

coubeatczech:
I also need this, let's have this in Opaleye!

### DIFF
--- a/src/Opaleye/Operators.hs
+++ b/src/Opaleye/Operators.hs
@@ -78,11 +78,5 @@ like = C.binOp HPQ.OpLike
 ors :: F.Foldable f => f (Column T.PGBool) -> Column T.PGBool
 ors = F.foldl' (.||) (T.pgBool False)
 
-ands :: F.Foldable f => f (Column T.PGBool) -> Column T.PGBool
-ands = F.foldl' (.&&) (T.pgBool True)
-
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Thanks for that. 

* Shouldn't `in_` and `notIn` be generalised to `Foldable`?
* Should it be `foldl'` not `foldr`?  The result can't be consumed lazily, so I would guess so.  On the other hand, maybe future potential optimisations could take advantage of laziness.  Anyway, it's probably not a big deal.
* Do we really need `notIn`?

coubeatczech:
* that's right, I added `Foldable` context
* ok, that's right, let's use strict fold
* I don't see why there shouldn't exist any case where `notIn` is needed, is it already expressable with something else?

tomjaguarpaw:
Sure, `notIn x y == not (in_ x y)`.

coubeatczech:
Now I fell a little stupid :-).
We also save chars, because `length "OR" < length "AND"` :-)

tomjaguarpaw:
No need to feel stupid :-)

tomjaguarpaw:
I will leave it up to you whether to keep it or remove it.  You made the patch so you get to decide.

bergmark:
I also suggested notIn :astonished:

tomjaguarpaw:
We may as well still have `ands` though!

Issue #68, opened 2015-07-13
Streaming interface for queries?


hvr:
I've already asked this for the `groundhog` library (see https://github.com/lykahb/groundhog/issues/41), and now I'm wondering whether `opaleye` has support or plans to support a streaming-like interface for querying many tuples from a database in constant-space.

tomjaguarpaw:
This is a very good idea.  [`postgresql-simple` already supports this](http://hackage.haskell.org/package/postgresql-simple-0.4.10.0/docs/Database-PostgreSQL-Simple.html#g:17) so it should be easy to add to Opaleye.

I have a feeling someone has already worked on such a patch, but a quick search didn't turn up anything.

We would need `foldWith_` which takes a `RowParser` rather than a `FromRow` constraint, but I'm sure Leon would be happy to merge that upstream.

tstat:
I submitted a PR that was accepted on this in May https://github.com/lpsmith/postgresql-simple/pull/147 and have opaleye functions for working with this locally, but have not made any PRs to opaleye yet as Leon has not published to hackage since Feburary.

tomjaguarpaw:
Thanks Travis!  Can you PR your patch anyway with a note saying we are waiting for an upstream Hackage release?  It would be good if people can see it's in the queue.

tstat:
Sure thing, I'll go ahead and submit it when I get a free moment.

tomjaguarpaw:
The PR is here https://github.com/tomjaguarpaw/haskell-opaleye/pull/69

tomjaguarpaw:
I'm happy to merge #69 and close #68 as soon as I'm convinced it's OK to force everyone to upgrade to the lastest postgresql-simple.  Feel free to convince me!

LukeHoersten:
Just bump the major version number. Back ported bug fixes can still be released under the previous major version.

tomjaguarpaw:
This has been merged in HEAD for some time.  Anyone who especially wants a Hackage release (and this applies in general) please open another issue.

PR #69, opened 2015-07-15
Add runQueryFold for streaming database queries

tstat:
**Note:** This will not work until [postgresql-simple commit e8562d6](https://github.com/tstat/postgresql-simple/commit/e8562d63c0807d32c70808a3aac289ce1d0f3e85) is published to hackage.

Because the where clause of `runQueryFoldExplicit` is a direct copy of `runQueryExplicit` I pulled out a function I named `prepareQuery`. I am not sure what module would be the best place for this function so any advice on that would be appreciated.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -33,7 +33,7 @@ library
     , case-insensitive    >= 1.2     && < 1.3
     , bytestring          >= 0.10    && < 0.11
     , contravariant       >= 1.2     && < 1.4
-    , postgresql-simple   >= 0.4.8.0 && < 0.6
+    , postgresql-simple   >= 0.5     && < 0.6
     , pretty              >= 1.1.1.0 && < 1.2
     , product-profunctors >= 0.6.2   && < 0.7
     , profunctors         >= 4.0     && < 5.2
--- a/src/Opaleye/RunQuery.hs
+++ b/src/Opaleye/RunQuery.hs
@@ -53,6 +53,30 @@ runQueryExplicit :: QueryRunner columns haskells
 runQueryExplicit qr conn q = PGS.queryWith_ parser conn sql
   where (sql, parser) = prepareQuery qr q
 
+-- | @runQueryFold@ streams the results of a query incrementally and consumes
+-- the results with a left fold.
+--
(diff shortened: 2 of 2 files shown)


tomjaguarpaw:
`foldWith_` has been pushed to Hackage so we are good to go.

tomjaguarpaw:
I'm happy to merge #69 and close #68 as soon as I'm convinced it's OK to force everyone to upgrade to the lastest postgresql-simple.  Feel free to convince me!

tstat:
Oh well, I'm fine either way, including this with a min version check (as is currently done in this PR) or bumping the postgresql-simple dependency. I noticed publishing this with the min version check was not mentioned as a possibility though, is it a poor solution for some reason?

tomjaguarpaw:
I failed to notice the min version check.  Yes, I'd rather not use CPP.

tstat:
Ok, I'll remove it from the PR, but for my own education why would you rather not use CPP?

tomjaguarpaw:
It's not strictly necessary in this case.  Forcing everyone to upgrade to
the latest postgresql-simple is (probably) fine so I'd rather not introduce
any extra potential brittleness.

tstat:
Ok, sounds good to me, thanks for the explanation. I pushed a new version that bumps the postgresql-simple version constraint.

tomjaguarpaw:
Right, it's about time I merged this very useful patch.  One question that needs settling before I do: isn't it strictly more general to use `MonadIO m` rather than `IO`?

tstat:
I would love to generalize `runQueryFold`, however, it is not possible with the current postgresql-simple library. The library postgresql-simple uses the accumulation function passed in within `bracket`, and in order to be generalized out of `IO` postgresql-simple would need to adopt one of the exception libraries that allows one to implement a generalized `bracket`. 

`MonadIO` is not one of these libraries as it does not enable generalizations of `IO` when it appears in the contravariant position, but there has been [talk](https://github.com/lpsmith/postgresql-simple/issues/9) of generalizing to `MonadBaseControl IO m` for years, so if that happened at some point we could do the same.

tomjaguarpaw:
OK, let's stick with what we have.

Issue #37, opened 2015-06-15
Extract CREATE statements from a table


ndmitchell:
Should be pretty easy to generate a default CREATE statement.

tomjaguarpaw:
This would be very useful.  It requires some thought.  When we create a table we have to tell the DBMS what types the columns have.  This means we need a mapping from (type level) PGTypes to value level encodings of these.  The latter don't exist yet, but I'll try to think of the most judicious way to add them.

coubeatczech:
One option, how to do this would be to use something similar to the `Writer`, with the difference in the `a` type, something like this:

    newtype SchemaMaker read dummy = SchemaMaker (PM.PackMap ColumnDescription () read ())

the mapping from the type level to the pg types should be pretty straightforward if we don't want to have the option of specifying the range of the pg type, if we are ok with one pg type per type on haskell type level.

Or do you have something that is more elegant in mind?

tomjaguarpaw:
I think I would prefer to have all the data you need to create the table already stored in the `Table` type.  This is quite easily doable, but is a breaking change since will need

    required :: String -> Writer (Column a) (Column a)

to change to

    required :: PGType a => String -> Writer (Column a) (Column a)

for some new class `PGType` which knows how to convert `PGText` into `"Text"` and `PGInt4` into `"Int4"` etc..

Your suggestion doesn't require a breaking change, however, since it only introduces new names.

coubeatczech:
So you mean, that the table properties would get one more field:

    data TableProperties writerColumns viewColumns =
      TableProperties (Writer writerColumns viewColumns) (View viewColumns) ColumnDescription

and the information, how to create a row in a `CREATE` statement would be stored there. 

Also, the user should be able to specify other attributes, that can't be taken from the type system like `not null`, that is indices, generation from sequence etc. We would probably need to add variants of `required` and `optional` with more arguments.

tomjaguarpaw:
> So you mean, that the table properties would get one more field

That's right.

> Also, the user should be able to specify other attributes, that can't be taken from the type system

Yes, and probably most importantly, keys and foreign keys.  However, it seems a bit awkward to do these.

coubeatczech:
How would that look like? I'm imagining, that the referencing of foreign key could work like `runUpdate`, the `required` would accept a `Table` and a function ` read -> Column a `. Although I'm not sure, if it would work for foreign keys aiming at a column in the same table.

[...]

PR #74, opened 2015-07-18
wip #37

coubeatczech:
So this is the draft of possible imlementation of #37.

Recap: The main problem with this issue is that changing the `Column` in read part of `TableProperties` to `TableColumn`, that enables having other information in the table than just the column name will possibly break type inference. Excerpt from `Schema.hs`:

    t1 :: Table T1 T1'
    t1 = Table "t1" $ p2 (T.optional "col0" Autogenerated, T.required "col1" NoIntOptions)

    query :: O.Query (Column PGInt8)
    query = proc () -> do
      (col1, col2) <- T.queryTable t1 -< ()
      returnA -< col1

Here, the `col2` is not specified so the type inference engine don't know which instance to use. So, one needs to provide explicit type signature.

    t1Query :: O.Query T1''
    t1Query = T.queryTable t1

    query' :: O.Query (Column PGInt8)
    query' = proc () -> do
      (col1, _) <- t1Query -< ()
      returnA -< col1

Then the `query'` compiles.

There are some solution ideas that come to mind:
* accept it, the user will need to provide type signatures
* somehow run a closed type family on the `read` part of `TableProperties`, that would replace all occurences of `TableColumn` with `Column`, that can be then provided to the inferencer. Or do the same thing with macro?
* leave the table definition as is, user needs to not modify `Column`s before doing schema operations.

Another obvious issue is that the type of table will change and there is one more table type for the user to specify. `T1` & `T'` are types provided in the `Table` definition. `T''` is a type that is used when helping the type inference.

    type T1 = (Maybe (Column PGInt8), Column PGInt8)
    type T1' = (TM.TableColumn PGInt8, TM.TableColumn PGInt8)
    type T1'' = (Column PGInt8, Column PGInt8)

Any ideas, how to do this in an elegant way?


tomjaguarpaw:
I have thought of something that might help with the types which is based on an idea I originally got from @bergmark.  We could have an (open) type function

    ColumnFromTableColumn (TableColumn a) (Column a)
    ColumnFromTableColumn (a, b) = (ColumnFromTableColumn a, ColumnFromTableColumn b)
    ColumnFromTableColumn (Foo a b) = Foo (ColumnFromTableColumn a) (ColumnFromTableColumn b)

which is generated automatically by `makeAdaptorAndInstance`.  In fact I hope a single type family could be generated that works in all cases just by switching a parameter in the same way that `Default` is parametrized.

You might want to give this a go.

coubeatczech:
Ok, I'll go in that direction and let's see.

domenkozar:
@tomjaguarpaw reducing boilerplate and getting started would help a lot, if one could create tables out of the schema that's a good improvement. Currently the alternative is to use opaleye-gen, which has a slight unmaintainable detaste.

PR #77, opened 2015-07-22
Expose QueryRunnerColumnDefault from RunQuery

bergmark:
We are using this in [our TH for generating column instances](https://github.com/silkapp/silk-opaleye/blob/master/src/Silk/Opaleye/TH/Column.hs)


tomjaguarpaw:
Thanks, you are right that this should be exposed.

PR #94, opened 2015-08-09
Add Semigroup Order instance

bergmark:


tomjaguarpaw:
I'm not against adding this, but is it actually useful?

bergmark:
It's useful to me because i tend to have `<>` from semigroups in scope.

tomjaguarpaw:
Fair enough.

bergmark:
Thanks!

Issue #98, opened 2015-08-14
Support for postgresql composite types


WraithM:
Postgresql supports [composite types](http://www.postgresql.org/docs/9.4/static/rowtypes.html). Composite types have this special string-literal syntax for constructing them. Here's a little example:

```sql
CREATE TYPE point AS (
    x double precision,
    y double precision
);

CREATE TABLE points (p point);
INSERT INTO points VALUES('(2.0, 3.0)');
```

It would be awesome if opaleye supported these types.

Right now, I'm hacking this in with `OtherLit` and a manual `FromField` instance with a custom parser:

```haskell
import Opaleye.Internal.HaskellDB.PrimQuery as HPQ
data PGPoint

pgPoint :: Point -> Column PGPoint
pgPoint (Point x y) = literalColumn . HPQ.OtherLit $
        "(" ++ show x ++ "," ++ show y ++ ")"

instance QueryRunnerColumnDefault PGPoint Point where
    queryRunnerColumnDefault = fieldQueryRunnerColumn

instance FromField Point where
     -- Custom parsing for the composite type goes here
```

I was thinking that an implementation of this could just add a `CompositeLit` constructor to the `Literal` type and a way of easily constructing `FromField` instances for composite types. For the `CompositeLit` part, I was thinking something like this:

```haskell
data Literal = ... | CompositeLit1 Literal | CompositeLit2 (Literal, Literal) | ...

defaultSqlLiteral :: SqlGenerator -> Literal -> String
defaultSqlLiteral g (CompositeLit1 l) = defaultSqlLiteral g l
defaultSqlLiteral g (CompositeLit2 (l1, l2)) =
    "(" ++ defaultSqlLiteral g l1 ++ "," ++ defaultSqlLiteral g l2 ++ ")"
...
```

Another possible implementation would be just to provide functions for constructing `OtherLit` in the style of composite types and functions for `FromField` instances.

tomjaguarpaw:
`CompositeLit [Literal]` is probably fine, just like argument application to functions.

WraithM:
Yeah, that was my first thought, but I guess I thought that explicitly making tuples would be more type-safe. Opaleye could just throw an error or something if the list is beyond a certain length or whatever could go wrong.

tomjaguarpaw:
You're right, it would be more typesafe.  However I think it's quite hard to make an AST typesafe.  The typesafety of the source syntax is enough, at least for now.

[...]

PR #99, opened 2015-08-19
Composite Types (Issue #98)

WraithM:
Adding composite types by creating a new data type Composite, which plays a similar role to Column. Created a new expression constructor for HaskellDB's PrimQuery, and modified HaskellDB's SqlColumn type.

Here's an example of using it:
```sql
create type location as (
    x double precision,
    y double precision
);

create table locations (
    id serial primary key,
    location1 location,
    location2 location
);
```

```haskell
{-# LANGUAGE EmptyDataDecls        #-}
{-# LANGUAGE FlexibleInstances     #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings     #-}
{-# LANGUAGE TemplateHaskell       #-}

module Main where

import           Data.Maybe                           (listToMaybe)
import           Data.Profunctor.Product.Default      as D
import           Data.Profunctor.Product.TH           (makeAdaptorAndInstance)

import qualified Database.PostgreSQL.Simple           as PSQL
import           Database.PostgreSQL.Simple.FromRow

import           Opaleye                              hiding (literalColumn)
import           Opaleye.Internal.Column
import           Opaleye.Internal.HaskellDB.PrimQuery as HPQ
import           Opaleye.Internal.PGTypes
import           Opaleye.Internal.RunQuery

data Location = Location
    { x :: Double
    , y :: Double
    } deriving (Show, Eq)

data PGLocation

pgLocation :: Location -> Composite PGLocation
pgLocation (Location x y) = WriteComposite . HPQ.ConstExpr . HPQ.OtherLit $
    "(" ++ show x ++ "::double precision," ++ show y ++ "::double precision)"

instance FromRow Location where
    fromRow = Location <$> field <*> field

instance D.Default QueryRunner (Composite PGLocation) Location where
    def = rowQueryRunner

data Locations = Locations
    { l1 :: Location
    , l2 :: Location
    } deriving (Show, Eq)

data LocationsP i a b = LocationsP
    { pId :: i
    , pl1 :: a
    , pl2 :: b
    } deriving (Show, Eq)

makeAdaptorAndInstance "pLocations" ''LocationsP

type WriteLocations = LocationsP
    (Maybe (Column PGInt4))
    (Composite PGLocation)
    (Composite PGLocation)

type ReadLocations = LocationsP
    (Column PGInt4)
    (Composite PGLocation)
    (Composite PGLocation)

locationsTable :: Table WriteLocations ReadLocations
locationsTable = Table "locations"
    (pLocations LocationsP
        { pId = optional "id"
        , pl1 = compositeRequired "location1" ["x", "y"]
        , pl2 = compositeRequired "location2" ["x", "y"]
        })

selectAllLocations :: Query ReadLocations
selectAllLocations = queryTable locationsTable

runLocationsQuery :: PSQL.Connection -> Query ReadLocations -> IO [(Int, Locations)]
runLocationsQuery c q = map mkLocations <$> runQuery c q
  where
    mkLocations :: LocationsP Int Location Location -> (Int, Locations)
    mkLocations (LocationsP i p1 p2) = (i, Locations p1 p2)

insertLocations :: PSQL.Connection -> Locations -> IO (Maybe Int)
insertLocations c (Locations p1 p2) =
    listToMaybe <$> runInsertReturning c locationsTable wlocations pId
  where
    wlocations = LocationsP Nothing (pgLocation p1) (pgLocation p2)

main :: IO ()
main = do
    c <- PSQL.connectPostgreSQL "dbname='testpoints' user='postgres'"

    printSql selectAllLocations

    _ <- insertLocations c (Locations (Location 1.2 2.1) (Location 2.3 62.1))
    _ <- insertLocations c (Locations (Location 3.2 6.1) (Location 2 51.1))
    runLocationsQuery c selectAllLocations >>= printLines
  where
    printSql = putStrLn . showSqlForPostgres
    printLines = putStrLn . unlines . map show
```

### DIFF
--- a/src/Opaleye/Column.hs
+++ b/src/Opaleye/Column.hs
@@ -2,9 +2,11 @@ module Opaleye.Column (module Opaleye.Column,
                        Column,
                        Nullable,
                        unsafeCoerce,
-                       unsafeCoerceColumn)  where
+                       unsafeCoerceColumn,
+                       unsafeCompositeField)  where
 
-import           Opaleye.Internal.Column (Column, Nullable, unsafeCoerce, unsafeCoerceColumn)
+import           Opaleye.Internal.Column (Column, Nullable, unsafeCoerce, unsafeCoerceColumn, 
+                                          unsafeCompositeField)
 import qualified Opaleye.Internal.Column as C
 import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 import qualified Opaleye.PGTypes as T
--- a/src/Opaleye/Internal/Column.hs
+++ b/src/Opaleye/Internal/Column.hs
@@ -19,6 +19,10 @@ unsafeCoerce = unsafeCoerceColumn
 unsafeCoerceColumn :: Column a -> Column b
 unsafeCoerceColumn (Column e) = Column e
 
+unsafeCompositeField :: Column a -> String -> Column b
+unsafeCompositeField (Column e) fieldName = 
+  Column (HPQ.CompositeExpr e fieldName)
+
 binOp :: HPQ.BinOp -> Column a -> Column b -> Column c
 binOp op (Column e) (Column e') = Column (HPQ.BinExpr op e e')
 
--- a/src/Opaleye/Internal/HaskellDB/PrimQuery.hs
+++ b/src/Opaleye/Internal/HaskellDB/PrimQuery.hs
@@ -17,6 +17,7 @@ data Symbol = Symbol String T.Tag deriving (Read, Show)
 
 data PrimExpr   = AttrExpr  Symbol
                 | BaseTableAttrExpr Attribute
+                | CompositeExpr     PrimExpr Attribute -- ^ Composite Type Query
                 | BinExpr   BinOp PrimExpr PrimExpr
                 | UnExpr    UnOp PrimExpr
                 | AggrExpr  AggrOp PrimExpr
--- a/src/Opaleye/Internal/HaskellDB/Sql.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql.hs
@@ -30,6 +30,7 @@ data SqlOrder = SqlOrder { sqlOrderDirection :: SqlOrderDirection
 
 -- | Expressions in SQL statements.
 data SqlExpr = ColumnSqlExpr  SqlColumn
+             | CompositeSqlExpr SqlExpr String
              | BinSqlExpr     String SqlExpr SqlExpr
              | PrefixSqlExpr  String SqlExpr
              | PostfixSqlExpr String SqlExpr
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -80,6 +80,7 @@ defaultSqlExpr gen expr =
     case expr of
       AttrExpr (Symbol a t) -> ColumnSqlExpr (SqlColumn (tagWith t a))
       BaseTableAttrExpr a -> ColumnSqlExpr (SqlColumn a)
+      CompositeExpr e x -> CompositeSqlExpr (defaultSqlExpr gen e) x
       BinExpr op e1 e2 ->
         let leftE = sqlExpr gen e1
             rightE = sqlExpr gen e2
--- a/src/Opaleye/Internal/HaskellDB/Sql/Print.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Print.hs
@@ -113,6 +113,7 @@ ppSqlExpr :: SqlExpr -> Doc
 ppSqlExpr expr =
     case expr of
       ColumnSqlExpr c     -> ppColumn c
+      CompositeSqlExpr s x -> parens (ppSqlExpr s) <> text "." <> text x
       ParensSqlExpr e -> parens (ppSqlExpr e)
       BinSqlExpr op e1 e2 -> ppSqlExpr e1 <+> text op <+> ppSqlExpr e2
       PrefixSqlExpr op e  -> text op <+> ppSqlExpr e
--- a/src/Opaleye/Internal/Table.hs
+++ b/src/Opaleye/Internal/Table.hs
@@ -88,6 +88,7 @@ runColumnMaker cm tag tableCols = PM.run (TM.runColumnMaker cm f tableCols) wher
   -- tablecols contained in the View (which would be naughty)
   mkName pe i = (++ i) $ case pe of
     HPQ.BaseTableAttrExpr columnName -> columnName
+    HPQ.CompositeExpr columnExpr fieldName -> mkName columnExpr i ++ fieldName
     _ -> "tablecolumn"
 
 runWriter :: Writer columns columns' -> columns -> [(HPQ.PrimExpr, String)]


WraithM:
And here's the result from that program:

```
       "location1x1_1" as "result2_2",
       "location1y2_1" as "result3_2",
       "location2x3_1" as "result4_2",
       "location2y4_1" as "result5_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   (location1).x as "location1x1_1",
                   (location1).y as "location1y2_1",
                   (location2).x as "location2x3_1",
                   (location2).y as "location2y4_1"
            FROM "locations" as "T1") as "T1") as "T1"
(1,Locations {l1 = Location {x = 1.2, y = 2.1}, l2 = Location {x = 2.3, y = 62.1}})
(2,Locations {l1 = Location {x = 3.2, y = 6.1}, l2 = Location {x = 2.0, y = 51.1}})
```

tomjaguarpaw:
Can you write a bit about why this is needed and in particular why `Colun` is not sufficient?

WraithM:
Sure! It basically comes down to the fact that `Column` assumes a 1-to-1 relationship with a column and a SQL expression (`PrimExpr`). It's not possible to express a query to a multiple rows with a single `PrimExpr` (at least with the way that the `HaskellDB` code is set up right now). With a row-type, I need multiple `PrimExpr`s to query the different fields, hence the `ReadComposite` constructor.

```haskell
newtype Column a = Column HPQ.PrimExpr deriving Show

data Composite a 
    = WriteComposite HPQ.PrimExpr
    | ReadComposite [HPQ.PrimExpr] deriving Show
```

One of my failed solutions to this problem actually tried to expand the definition of `PrimExpr` so that it could express queries to multiple fields, but it became really difficult really fast. A possible solution would be to redefine the `Column` type to be structurally the same as the `Composite` type if you don't like the idea of multiple types.

Also, the `Default ColumnMaker (Column a) (Column a)` is too restrictive. I need to be able to carry around multiple `PrimExpr`s around as I do with the `composite` function.

```haskell
column :: ColumnMaker (C.Column a) (C.Column a)
column = ColumnMaker
         (PM.PackMap (\f (IC.Column s)
                      -> fmap IC.Column (f s)))

composite :: ColumnMaker (C.Composite a) (C.Composite a)
composite = ColumnMaker (PM.PackMap pm)
  where
    pm f (IC.WriteComposite x) = fmap IC.WriteComposite (f x)
    pm f (IC.ReadComposite xs) = fmap IC.ReadComposite (T.traverse f xs)

instance Default ColumnMaker (C.Column a) (C.Column a) where
   def = column

instance Default ColumnMaker (C.Composite a) (C.Composite a) where
  def = composite
```

Does that answer your question?

WraithM:
I would also add that `Composite` can now expose a different interface than `Column`. Row-types are different from columns in postgresql, so I think they should be expressed with different types. Row-types are actually more similar to a table than a column in many respects.

LukeHoersten:
+1

tomjaguarpaw:
What's the difference between `ReadComposite` and `WriteComposite`?

tomjaguarpaw:
Can you say what went wrong when you tried to extend `PrimExpr` to deal with multiple fields?  This really seems like the right solution.

tomjaguarpaw:
Concretely it seems like there should be two extra constructors to `PrimExpr`

    | Row [PrimExpr]
    | RowField PrimExpr String

which do `ROW(foo, bar, baz)` and `myrow.xyz` respectively.

WraithM:
> What's the difference between `ReadComposite` and `WriteComposite`?

`ReadComposite` is for querying composite types:

```haskell
compositeRequired :: String -> [String] -> TableProperties (Composite a) (Composite a)
compositeRequired columnName fields = T.TableProperties 
    (T.requiredComposite columnName)
    (View . ReadComposite $ map (HPQ.CompositeTableAttrExpr columnName) fields)
```

`WriteComposite` is for inserting (referring to my example code above rather than my PR):

```haskell
pgLocation :: Location -> Composite PGLocation
pgLocation (Location x y) = WriteComposite . HPQ.ConstExpr . HPQ.OtherLit $
    "(" ++ show x ++ "::double precision," ++ show y ++ "::double precision)"
```

> Can you say what went wrong when you tried to extend `PrimExpr` to deal with multiple fields? That really seems like the right solution.

What you wrote (with `Row` and `RowField`) is basically exactly what my failed attempt was.

There are lots of places where the code assumes that there's one `PrimExpr` for a select attribute. For example in `Opaleye.Internal.Sql` (relevant snippet included below), you'd have to flatten down the `Row` into a single list.

```haskell
sql :: ([HPQ.PrimExpr], PQ.PrimQuery, T.Tag) -> Select
sql (pes, pq, t) = SelectFrom $ newSelect { attrs = SelectAttrs (ensureColumns (makeAttrs pes))
                                          , tables = [pqSelect] }
  where pqSelect = PQ.foldPrimQuery sqlQueryGenerator pq
        makeAttrs = flip (zipWith makeAttr) [1..]
        makeAttr pe i = sqlBinding (Symbol ("result" ++ show (i :: Int)) t, pe)

baseTable :: String -> [(Symbol, HPQ.PrimExpr)] -> Select
baseTable name columns = SelectFrom $
    newSelect { attrs = SelectAttrs (ensureColumns (map sqlBinding columns))
              , tables = [Table (HSql.SqlTable name)] }

sqlBinding :: (Symbol, HPQ.PrimExpr) -> (HSql.SqlExpr, Maybe HSql.SqlColumn)
sqlBinding (Symbol sym t, pe) =
  (sqlExpr pe, Just (HSql.SqlColumn (T.tagWith t sym)))
```

Changing this complicates scope management for the field aliases (`AS` names) for select attributes. Basically, I couldn't get field aliases to match up with the appropriate `PrimExpr`. It might be possible, but it seems like it would significantly complicate the `HaskellDB` backend.

Opaleye supports this paradigm of having multiple types for different selections rather than just `Column`. I think that there really should be a different interface for composite-types rather than reusing the `Column` type. What are your thoughts on that?

tomjaguarpaw:
> I think that there really should be a different interface for composite-types rather than reusing the Column type. What are your thoughts on that?

I am very much against treating composite types any differently, if we can possibly avoid it.  I still don't fully understand what goes wrong with my suggestion (and your earlier attempt).  Why can't an SQL field name refer to an entire composite, rather than breaking it up?  I'm imagining generating code like

    SELECT (ROW(1, 5.5, "Hello") :: mycompositetype) AS "result1"
    ....

and

    SELECT compositetypefield.x AS "result1"
    ...

Is that what you were thinking of too?

WraithM:
I'm not sure I understand the `SELECT (ROW(1, 5.5, "Hello") :: mycompositetype) AS "result1"` part. The `ROW` syntax is for inserts/updates. However, for inserts/updates, that might be a good addition to the `PrimExpr` type rather than using a string literal. I think I may have misunderstood what you meant by the `Row` constructor in my response above to your question about it.

For `SELECT compositetypefield.x AS "result1"`, I actually did modify `PrimExpr` to support that:

```haskell
                | CompositeTableAttrExpr Attribute Attribute -- ^ Composite Type Query
```

We could rename `CompositeTableAttrExpr` to `RowField` and give it more structure.

The problem comes when you want to select multiple fields for a composite type in Opaleye's frontend. That's why we need a new type. Let me construct an example.

Say we only had `Column`. In order to query my multiple fields in a composite type, I'd have to include each field of the composite type individually in the table definition. Instead of what I wrote in my example at the top, you'd write something like:

```haskell
data LocationsP i a b c d = LocationsP
    { pId :: i
    , pl1x :: a
    , pl1y :: b
    , pl2x :: c
    , pl2y :: d
    } deriving (Show, Eq)

locationsTable :: Table WriteLocations ReadLocations
locationsTable = Table "locations"
    (pLocations LocationsP
        { pId = optional "id"
        , pl1x = compositeRequired "location1" "x" 
        -- compositeRequired now takes a column name of the composite type and a field name, 
        -- instead of all of the subfields that you want to query like I have above
        , pl1y = compositeRequired "location1" "y"
        , pl2x = compositeRequired "location2" "x"
        , pl2y = compositeRequired "location2" "y"
        })
```

There's a combinatoric problem where if you have many subfields, you have to include a lot of `Column`s to select each one individually because you need a multiple `PrimExpr`s to express the idea that you want to select multiple rows. `Column` only supports one `PrimExpr`, but if I want to do what I did in my initial example, we need a new type or radically change the HaskellDB backend. Of course, we could just add `CompositeTableAttrExpr`/`RowField` to Opaleye's backend, and then I could build whatever I want for composite types in a separate library. However, I think that it'd be nice to have support for composite types in the main Opaleye code.

WraithM:
Perhaps `Composite` should be renamed to `Row` because what it's really doing underneath is using `FromRow` instances. There may be a way to conceptually rethink this as handling `FromRow` more generally, rather than just solving my specific problem of composite-types. The fact that the `ReadComposite` constructor has a list is critical to this functioning properly.

```haskell
data Composite a 
    = WriteComposite HPQ.PrimExpr
    | ReadComposite [HPQ.PrimExpr] deriving Show

rowQueryRunner :: FromRow haskell => QueryRunner (Composite a) haskell
rowQueryRunner = QueryRunner (P.rmap (const ()) U.unpackspecComposite) (const fromRow) (const True)
```

tomjaguarpaw:
Why could we not have the following setup:

    type WriteLocations = LocationsP
        (Maybe (Column PGInt4))
        (Column PGLocation)
        (Column PGLocation)
    
    type ReadLocations = LocationsP
        (Column PGInt4)
        (Column PGLocation)
        (Column PGLocation)
    
    locationsTable :: Table WriteLocations ReadLocations
    locationsTable = Table "locations"
        (pLocations LocationsP
            { pId = optional "id"
            , pl1 = required "location1"
            , pl2 = required "location2"
            }

    -- These to be filled in with some simple definitions
    locationX, locationY :: Column PGLocation -> Column PGFloat8
    location :: Column PGFloat8 -> Column PGFloat8 -> Column PGLocation
    pgLocation :: Location -> Column PGLocation
    instance Default QueryRunner (Column PGLocation) Location where

WraithM:
Where would the definition of the subfield names go? ("x" and "y") I think you still need something like `compositeRequired` for the `Table`, and the definition of `Column` would need to be modified to be able to support `FromRow` types.

```haskell
data Column a
    = Column HPQ.PrimExpr
    | Row [HPQ.PrimExpr]
```

`Row` for allowing support for `FromRow` types.

tomjaguarpaw:
The subfield names would appear in `locationX` and `locationY`, and maybe also in the `QueryRunner` instance, though it should be possible to remove any duplication.

Why would `Column` need to be modified to support `FromRow` instances?  You can already write an

    instance Default QueryRunner (Column PGLocation) Location where

using `FromRow` if you want.

WraithM:
Could you give me an example? I couldn't make a `instance Default QueryRunner (Column PGLocation) Location where` using `FromRow`.

tomjaguarpaw:
You don't (necessarily) need to use `FromRow`.  You can use the (hypothetical) `locationX` and `locationY` and pack the fields into a `Location` on the other side.

What I'm suggesting here is doing it "by hand".  Ideally there would be some generic way of doing this with composite types, but that generic way doesn't actually need anything new within Opaleye.

tomjaguarpaw:
To be concrete, I think this should do it

    queryRunnerLocation :: QueryRunner (Column PGLocation) Location
    queryRunnerLocation = dimap (locationX &&& locationY) (uncurry Location) default

WraithM:
What's `default` here?

WraithM:
Thank you for humoring me with this discussion :+1:

tomjaguarpaw:
`Data.Profunctor.Product.Default.def :: QueryRunner (Column PGFloat8, Column PGFloat8) (Double, Double)`

Very happy to have this discussion and improve Opaleye with composite types!

WraithM:
Very interesting! That should work if `locationX` and `locationY` are easy to make. I feel like `locationX` and `locationY` might be tricky. What do you think those would look like? Would they be implemented in terms of finding the field in your `Row :: [PrimExpr] -> PrimExpr` constructor? Doesn't this involve creating parsers for the row-type literal syntax?

tomjaguarpaw:
In fact I don't see why you can't write a suitable instance using `FromRow`.  `QueryRunner` is the following

data QueryRunner columns haskells =
  QueryRunner (U.Unpackspec columns ())
              (columns -> RowParser haskells)
              (columns -> Bool)

so you just need something like

    queryRunnerLocation :: QueryRunner Column PGLocation) Location
    queryRunnerLocation = QueryRunner (P.rmap (const ()) (const fromRow) (const True)

(the first argument is just the same as the one from `fieldQueryRunnerColumn`)

tomjaguarpaw:
They would need the extra `RowField` constructor described above.  I would try my newest suggestion first before diving in with anything else.

WraithM:
So, that was my very very first attempt. The problem is that you need an `Unpackspec`. The default `Unpackspec` for columns depends on the definition of `Column` being one `PrimExpr`.

```haskell
unpackspecColumn :: Unpackspec (C.Column a) (C.Column a)
unpackspecColumn = Unpackspec
                   (PM.PackMap (\f (IC.Column pe) -> fmap IC.Column (f pe)))

unpackspecComposite :: Unpackspec (C.Composite a) (C.Composite a)
unpackspecComposite = Unpackspec (PM.PackMap pm)
  where
    pm :: Applicative f => (HPQ.PrimExpr -> f HPQ.PrimExpr) -> C.Composite a -> f (C.Composite a)
    pm f (IC.WriteComposite x) = fmap IC.WriteComposite (f x)
    pm f (IC.ReadComposite xs) = fmap IC.ReadComposite (T.traverse f xs)
```

tomjaguarpaw:
So maybe I'm misunderstanding what the `RowParser` does for composite types.  Can't you say

    SELECT mylocation FROM ...

and parse it using the `RowParser`?  Are you saying that instead you have to

    SELECT mylocation.x, mylocation.y FROM ...

?

WraithM:
No. You can't. I am saying that you have to use the latter query to use `FromRow` instances.

In that first case, if you do `SELECT mylocation FROM`, you get out a special literal syntax. You get one column out, and therefore you must use a `FieldParser`. `FromField` instances for that special literal syntax don't exist in postgresql-simple. In the second case, `SELECT (location).x, (location).y FROM`, you get out two columns, and then you can use a `RowParser`. (The parens are necessary for not confusing the SQL parser)

WraithM:
A third option is to use `SELECT (location).* FROM`, I believe. You do get out two columns in that case, but the problem there is you run into naming problems. You cannot write `SELECT (location).* AS "result1" FROM`, that produces a SQL error.

tomjaguarpaw:
OK, sorry for getting the wrong end of the stick.  I didn't closely follow all the above discussion of `FromRow` with composite types.

In that case I suggest you try my first proposal.  It will require adding the `RowField` constructor but beyond that there should be no changes needed to the internals.

tomjaguarpaw:
Yeah I wouldn't use `*`.  That sounds likely to cause more problems.

WraithM:
So, my PR here has `RowField` (it's just called `CompositeTableAttrExpr`, bad naming on my part).

tomjaguarpaw:
So it does.  I don't have strong opinions on the name.  I'll leave it more-or-less up to you.  However, instead of

    CompositeTableAttrExpr Attribute Attribute

it should be

    CompositeTableAttrExpr PrimExpr Attribute

tomjaguarpaw:
You'll probably want to write some "`unsafe`" functions to access fields and then I think you're set!

WraithM:
Yeah, I can add more structure to `CompositeTableAttrExpr`!

What do you mean by unsafe functions? Accessing which fields?

tomjaguarpaw:
Probably just `unsafeCompositeField :: Column a -> String -> Column b`

WraithM:
Would you like me to change the definition of `Column` to:

```haskell
data Column a = Column HPQ.PrimExpr | Row [HPQ.PrimExpr]
```

or should I leave the `Composite` type? I might change the name from `Composite` to `Row`.

tomjaguarpaw:
Doesn't my suggestion remove the need for `Composite`?

WraithM:
I'm sorry for the confusion on my part. If I change the definition of `Column`, then yes, but the problem of `Unpackspec` still stands. You can't use a `RowParser` with `Column` because of the way that `Unpackspec` works for `Column`.

tomjaguarpaw:
But you don't need to use `Unpackspec` anymore!  All you need is

    queryRunnerLocation = dimap (locationX &&& locationY) (uncurry Location) Default.def

and if you want to use that in the `Default` instance

    instance Default QueryRunner (Column PGLocation) Location where
        def = queryRunnerLocation

tomjaguarpaw:
Here's some code that does the job.  You just need to fill in `unsafeCompositeField`.

````haskell
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}

import Opaleye
import Data.Profunctor.Product.Default
import Data.Profunctor
import Control.Arrow

data Location = Location
    { x :: Double
    , y :: Double
    } deriving (Show, Eq)

data PGLocation

locationX, locationY :: Column PGLocation -> Column PGFloat8
locationX = flip unsafeCompositeField "x"
locationY = flip unsafeCompositeField "y"

unsafeCompositeField :: Column a -> String -> Column b
unsafeCompositeField = undefined

queryRunnerLocation = dimap (locationX &&& locationY) (uncurry Location) def

instance Default QueryRunner (Column PGLocation) Location where
    def = queryRunnerLocation
````

**Comment by WraithM** [comment]
I think I'm having some misunderstandings of `Profunctor`s. Does `def :: QueryRunner (Column PGFloat8, Column PGFloat8) (Double, Double)` use a RowParser? How does that work?

**Comment by tomjaguarpaw** [comment]
`def :: QueryRunner (Column PGFloat8, Column PGFloat8) (Double, Double)` constructs a `QueryRunner` which contains (amongst other bits) a `RowParser (Double, Double)`.

**Comment by WraithM** [comment]
I can't get the `instance Default QueryRunner (Column PGLocation) Location where` solution to work. The problem is with the definition of the `Table`:

```haskell
required :: String -> TableProperties (Column a) (Column a)
required columnName = T.TableProperties
  (T.required columnName)
  (View (Column (HPQ.BaseTableAttrExpr columnName)))
```

There's no way to define a `TableProperties` for `Column`s that is able to query multiple columns on the backend. Therefore a `Table` definition as such:

```haskell
locationsTable :: Table WriteLocations ReadLocations
locationsTable = Table "locations"
    (pLocations LocationsP
        { pId = optional "id"
        , pl1 = required "location1" -- Can't query multiple columns on the backend
        , pl2 = required "location2"
        }
```

will always generate queries like this:

```sql
SELECT "id" as "id0_1",
                   "location1" as "location11_1",
                   "location2" as "location22_1"
            FROM "locations" as "T1"
```

when I really need it to generate this:

```sql
SELECT "id" as "id0_1",
                   (location1).x as "location1x1_1",
                   (location1).y as "location1y2_1",
                   (location2).x as "location2x3_1",
                   (location2).y as "location2y4_1"
            FROM "locations" as "T1"
```

I tried to make a new version of `required` (analogous to my `compositeRequired` only for `Column`s instead of `Composite`) that would generate the correct queries, and I can't figure it out without changing the definition of `Column`. I'm pretty sure it's not possible. The only way to make a correct `Table` definition is with a new type or changing the definition of `Column`.

tomjaguarpaw:
Show me your definition of `unsafeCompositeField`.

tomjaguarpaw:
I can see there's a problem with the `QueryRunner` instance, because the typeclass mechanism assumes it comes from a `QueryRunnerColumnDefault`.  Until that can be resolved just use the explicit version.  Here's complete code, except `unsafeCompositeField` that you have to fill in yourself.

````haskell
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TemplateHaskell #-}

import Opaleye
import Data.Profunctor.Product.Default
import Data.Profunctor.Product.TH
import Data.Profunctor.Product
import Data.Profunctor
import Control.Arrow

data Location = Location
    { x :: Double
    , y :: Double
    } deriving (Show, Eq)

data PGLocation

type WriteLocations = LocationsP
    (Maybe (Column PGInt4))
    (Column PGLocation)
    (Column PGLocation)

type ReadLocations = LocationsP
    (Column PGInt4)
    (Column PGLocation)
    (Column PGLocation)

data Locations = Locations
    { l1 :: Location
    , l2 :: Location
    } deriving (Show, Eq)

data LocationsP i a b = LocationsP
    { pId :: i
    , pl1 :: a
    , pl2 :: b
    } deriving (Show, Eq)

makeAdaptorAndInstance "pLocations" ''LocationsP

locationsTable :: Table WriteLocations ReadLocations
locationsTable = Table "locations"
    (pLocations LocationsP
        { pId = optional "id"
        , pl1 = required "location1"
        , pl2 = required "location2"
        })

locationX, locationY :: Column PGLocation -> Column PGFloat8
locationX = flip unsafeCompositeField "x"
locationY = flip unsafeCompositeField "y"

unsafeCompositeField :: Column a -> String -> Column b
unsafeCompositeField = error "You need to implement unsafeCompositeField!"

queryRunnerLocation :: QueryRunner (Column PGLocation) Location
queryRunnerLocation = dimap (locationX &&& locationY) (uncurry Location) def

unpackspecLocation :: Unpackspec (Column PGLocation) ()
unpackspecLocation = dimap (locationX &&& locationY)
                           (const () :: (Column PGFloat8, Column PGFloat8) -> ())
                           def

instance Default QueryRunner (Column PGLocation) Location where
    def = queryRunnerLocation

instance Default Unpackspec (Column PGLocation) () where
    def = unpackspecLocation

printSql = do
  putStrLn $ showSqlForPostgresExplicit
                (p2 (unpackspecLocation, unpackspecLocation))
                (fmap (\(LocationsP _ a b) -> (a, b)) (queryTable locationsTable))

queryDB conn = do
  r <- runQueryExplicit
         (p2 (queryRunnerLocation, queryRunnerLocation))
         conn
         (fmap (\(LocationsP _ a b) -> (a, b)) (queryTable locationsTable))

  let _ = r :: [(Location, Location)]

  print r
````

**Comment by WraithM** [comment]
I think you didn't understand my argument in my last comment. The problem happens before  `unsafeCompositeField` is called. The problem is with the definition of `locationsTable`. `queryTable locationsTable` generates the wrong query. The problem with the definition of `locationsTable` comes down to the definition of `required`. The generated query from `required` is `Column (HPQ.BaseTableAttrExpr columnName)`. `BaseTableAttrExpr columnName` compiles down to `"columnName"`. This is not correct. It needs to query the different subfields of the column, even if you can parse the rows of the output, which your solution does do. The definition of `Column` does not support creating the correct query. Your code does create a correct `RowParser`, but that doesn't matter because `required` can't generate the correct query because of the definition of `Column`.

I took your code and tried to highlight exactly what's going wrong: http://lpaste.net/1391113841017880576

The output from this example is this:
```sql
       "location11_1" as "result2_2",
       "location22_1" as "result3_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   "location1" as "location11_1",
                   "location2" as "location22_1"
            FROM "locations" as "T1") as "T1") as "T1"
opaleyecompositetest: You need to implement unsafeCompositeField!
```

`queryTable locationsTable` needs to generate this instead:

```sql
       "location1x1_1" as "result2_2",
       "location1y2_1" as "result2_2",
       "location2x1_1" as "result2_2",
       "location2y2_1" as "result3_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   (location1).x as "location1x1_1",
                   (location1).y as "location1y2_1",
                   (location2).x as "location2x1_1",
                   (location2).x as "location2y2_1"
            FROM "locations" as "T1") as "T1") as "T1"
```

To reiterate, the definition of `Column` does not have the ability to generate the correct query.

WraithM:
I should also add that you can't make a correct definition of `unsafeCompositeField` either for exactly the same reason.

WraithM:
Okay, I might have lied. I think I might have made a definition of `unsafeCompositeField` that works.

tomjaguarpaw:
> `"BaseTableAttrExpr columnName` compiles down to `"columnName"`. This is not correct.

I think it is correct :)

WraithM:
Yeah, you might be right. I'm working on it. Hold up. It requires some changes to the backend.

WraithM:
Yeah, I'm sorry for doubting you. You're totally right.

This:
```haskell
unsafeCompositeField :: Column a -> String -> Column b
unsafeCompositeField (Column e) fieldName = Column (CompositeTableAttrExpr e fieldName)
```

Gives you:

```sql
SELECT ("location11_1").x as "result1_2",
       ("location11_1").y as "result2_2",
       ("location22_1").x as "result3_2",
       ("location22_1").y as "result4_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   "location1" as "location11_1",
                   "location2" as "location22_1"
            FROM "locations" as "T1") as "T1") as "T1"
```

`queryDB` gives me:

```haskell
[(Location {x = 1.2, y = 2.1},Location {x = 2.3, y = 62.1}),(Location {x = 3.2, y = 6.1},Location {x = 2.0, y = 51.1}),(Location {x = 1.2, y = 2.1},Location {x = 2.3, y = 62.1}),(Location {x = 3.2, y = 6.1},Location {x = 2.0, y = 51.1})]
```

tomjaguarpaw:
Fantastic!  Good work.

WraithM:
```haskell
{-# LANGUAGE FlexibleInstances     #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings     #-}
{-# LANGUAGE TemplateHaskell       #-}

import           Control.Arrow

import           Data.Profunctor
import           Data.Profunctor.Product
import           Data.Profunctor.Product.Default
import           Data.Profunctor.Product.TH

import qualified Database.PostgreSQL.Simple           as PSQL
import           Opaleye
import           Opaleye.Internal.Column              (Column (Column))
import           Opaleye.Internal.HaskellDB.PrimQuery as HPQ

data Location = Location
    { x :: Double
    , y :: Double
    } deriving (Show, Eq)

data PGLocation

type WriteLocations = LocationsP
    (Maybe (Column PGInt4))
    (Column PGLocation)
    (Column PGLocation)

type ReadLocations = LocationsP
    (Column PGInt4)
    (Column PGLocation)
    (Column PGLocation)

data Locations = Locations
    { l1 :: Location
    , l2 :: Location
    } deriving (Show, Eq)

data LocationsP i a b = LocationsP
    { pId :: i
    , pl1 :: a
    , pl2 :: b
    } deriving (Show, Eq)

makeAdaptorAndInstance "pLocations" ''LocationsP

locationsTable :: Table WriteLocations ReadLocations
locationsTable = Table "locations"
    (pLocations LocationsP
        { pId = optional "id"
        , pl1 = required "location1"
        , pl2 = required "location2"
        })

locationX, locationY :: Column PGLocation -> Column PGFloat8
locationX = flip unsafeCompositeField "x"
locationY = flip unsafeCompositeField "y"

unsafeCompositeField :: Column a -> String -> Column b
unsafeCompositeField (Column e) fieldName = Column (CompositeTableAttrExpr e fieldName)

queryRunnerLocation :: QueryRunner (Column PGLocation) Location
queryRunnerLocation = dimap (locationX &&& locationY) (uncurry Location) def

unpackspecLocation :: Unpackspec (Column PGLocation) ()
unpackspecLocation = dimap (locationX &&& locationY)
                           (const () :: (Column PGFloat8, Column PGFloat8) -> ())
                           def

instance Default Unpackspec (Column PGLocation) () where
    def = unpackspecLocation

printSql = do
  putStrLn $ showSqlForPostgresExplicit
                (p2 (unpackspecLocation, unpackspecLocation))
                (fmap (\(LocationsP _ a b) -> (a, b)) (queryTable locationsTable))

queryDB conn = do
    r <- runQueryExplicit
            (p2 (queryRunnerLocation, queryRunnerLocation))
            conn
            (fmap (\(LocationsP _ a b) -> (a, b)) (queryTable locationsTable))

    let _ = r :: [(Location, Location)]

    print r

main :: IO ()
main = do
    c <- PSQL.connectPostgreSQL "dbname='testpoints' user='postgres'"

    printSql

    queryDB c
```

tomjaguarpaw:
What I'd like from you now is a PR with just the `CompositeTableAttrExpr` relevant stuff (including `unsafeCompositeField`).  I can merge that and then we can get on with whatever comes next (creating `ROW`s?).

WraithM:
Sure. Sounds good.

Just for my curiosity, why are you opposed to a new type? This solution feels like there's more boilerplate for the user than this original PR.

tomjaguarpaw:
We'll be able to clear up the boilerplate pretty easily.  I'm against a new type because it doesn't coherently fit in with the overall design on the library.

tomjaguarpaw:
Anyone is more than welcome to come up with a new type *outside* the library and wrap `Column` etc.. I just don't want it to be treated specially internally.

WraithM:
Makes sense. I might actually end up building my own `Row` type, and I can do that freely with this solution. Thank you for your perspective!

tomjaguarpaw:
You're welcome, and well done getting the internals all working.  There are not many comments in there so you managed to understand it very well.  I'm impressed.

WraithM:
Okay, I updated the PR.

WraithM:
I rebased and did evil git magic to get rid of the commit from when I initially made this PR. This PR is now only the `CompositeTableAttrExpr` and `unsafeCompositeField` stuff.

tomjaguarpaw:
Shouldn't `CompositeTableAttrExpr` be called `CompositeExpr`?  It doesn't appear, generally, in a table.

Also `SqlComposite` shouldn't be a constructor of `SqlColumn`.  It should be in `SqlExpr`, called `CompositeSqlExpr`.  That way we wouldn't have the strange-looking `ppColumn (SqlComposite s x) = parens (ppSqlExpr s) <> text "." <> text x` which I think serves no purpose.

tomjaguarpaw:
But other than that looks good, thanks!  If you make those changes I will merge.

WraithM:
I called it "CompositeTableAttrExpr" because the standard column attribute expression was "BaseTableAttrExpr", but maybe I misunderstood what that meant. I like `CompositeExpr` more :)

You're totally right about `SqlComposite`.

I've made those changes!

tomjaguarpaw:
Looks perfect.  Can you squash the commits into one again and then I'll merge.

WraithM:
Done.

hesselink:
What is the status of this? There is a pull request that's been merged, but I couldn't find an easy description of how to use it, and this ticket is still open.

I'm currently trying to return an array of pairs (basically a `Map`) from a query, and I'm wondering if the current infrastructure is enough for that, or if I need to go another route. I looked at the sources a bit, and it seems like opaleye has `ListExpr`, which seems to be what I need. But that needs a `FromField (a, b)` instance in postgresql-simple, which doesn't exist and isn't as easy to write as I was hoping.

tomjaguarpaw:
I'm afraid I don't know.  @WraithM is probably best placed to answer this.

lippling:
This is interesting to me as well.

ForestPhoenix:
Could somebody explain me why row types are not just columns of columns?

```
CREATE TYPE mynames AS (
	first_name text,
	last_name text
);

CREATE TABLE mynames_table ( twonames mynames );
INSERT INTO mynames_table VALUES (('first', 'last')::mynames)
```

This works, so why go through all the hassle with unsafeCompositeFiled?
Is it a limitation in postgresql-simple?

tomjaguarpaw:
Hi ForestPhoenix.  I'm not sure what you're really asking.  As far as I recall, the idea is that you define

    data MyNames
    firstName :: Column MyNames -> Column PGText
    firstName c = unsafeCompositeField c "first_name"

    lastName :: Column MyNames -> Column PGText
    lastName c = unsafeCompositeField c "last_name"

Did you have an alternative in mind?

pranaysashank:
@tomjaguarpaw Is there a way to write a `DefaultFromField PGMyNames MyNames` once we have both
```haskell
firstName :: Column PGMyNames -> Column PGText
lastName :: Column PGMyNames -> Column PGText
```
Based on this code snippet https://github.com/tomjaguarpaw/haskell-opaleye/pull/99#issuecomment-133537383, it is trivial to write a `Default FromFields (Column PGMyNames) MyNames` (function `queryRunnerLocation`). 

The problem is this overlaps with the instance `DefaultFromField a b => Default FromFields (Column a) b` from the library. Any idea how we can get around this without having to write a custom FieldParser?

tomjaguarpaw:
You can get around the overlapping instance problem with

```haskell
data Wrapped a = Wrapped a

instance Default FromFields (Wrapped (Column PGMyNames)) MyNames where
    def = ...
```

which isn't particularly pleasant, but will work. It means you'll have to wrap the `Column PGMyNames` whenever you `runSelect`, which is a bit sad. Failing that I think the only other approach would be getting better composite type support in Opaleye via getting better composite type support in postgresql-simple.

pranaysashank:
Thanks for the swift reply! 

I suppose I can use a `newtype Wrapped`  and use wrapped columns even in TableFields definition after writing `Default ToFields MyNames (Column PGMyNames))`.

And question I had was about writing composite types to the Db, currently I'm just doing

```haskell

data MyNames = MyNames {firstName :: Text, lastName :: Text}

instance Default ToFields MyNames (Column PGMyNames) where
    def = dimap myNamesToText (unsafeCast "mynames") (def :: ToFields Text (Column SqlText))
      where
        myNamesToText :: MyNames -> Text
        myNamesToText MyNames {..} = "(" <> show firstName <> "," <> show lastName <> ")"
```
But given that we already have `ToFields Text SqlText` can we use it to write these instances in a better way, like how we can for tables?

tomjaguarpaw:
I'm not sure what you mean.  You *are* using `ToFields Text (Column SqlText))`. What did you mean by "a better way"?

pranaysashank:
Ah yes, I am using `def :: ToFields Text (Column SqlText)` but that is to convert result of `myNamesToText`. I am wondering if there's a better way to write the function instead of using `show firstName` & `show lastName`  i.e. to use the `ToFields` instances of the constituent fields of the composite type (`MyNames`) to write the `ToFields MyNames (Column PGMyNames)`

tomjaguarpaw:
I see. The `ToFields` instances will give you `Column Text`s. I can't remember whether there's a function for turning `(Column Text, Column Text)` into a composite! Maybe there is or maybe it could be written easily but my memory of that is hazy.

tomjaguarpaw:
After a quick look through the codebase I can see that `CompositeExpr` gives a deconstruct composites, but there is currently no way to construct them.

PR #102, opened 2015-08-21
Support table schemas other than PostgreSQL's default "public"

k0001:
This is a backwards incompatible change that adds another parameter to the `Table` constructor so that the schema where the table is found can be specified.

http://www.postgresql.org/docs/9.3/static/ddl-schemas.html

### DIFF
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -64,6 +64,10 @@ manipulation tutorial you can see an example of when they might differ.
 >                                       , required "age"
 >                                       , required "address" ))
 
+By default, the table `"personTable"` is looked up in PostgreSQL's
+default `"public"` schema. If we wanted to specify a different schema we
+could have used the `TableWithSchema` constructor instead of `Table`.
+
 To query a table we use `queryTable`.
 
 (Here and in a few other places in Opaleye there is some typeclass
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -137,7 +137,7 @@ table4 = twoIntTable "table4"
 
 table5 :: O.Table (Maybe (Column O.PGInt4), Maybe (Column  O.PGInt4))
                   (Column O.PGInt4, Column O.PGInt4)
-table5 = O.Table "table5" (PP.p2 (O.optional "column1", O.optional "column2"))
+table5 = O.TableWithSchema "public" "table5" (PP.p2 (O.optional "column1", O.optional "column2"))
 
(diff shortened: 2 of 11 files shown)


tomjaguarpaw:
I'm happy to add this but it needs to be kept backwards compatible.  Perhaps add a separate constructor `TableWithSchema`?

k0001:
Ah, excellent idea, I'll do that.
Il 22/ago/2015 05:17 AM, "tomjaguarpaw" <notifications@github.com> ha
scritto:

k0001:
Ok, in https://github.com/k0001/haskell-opaleye/commit/921768e115a340376a121408ad4b08a94573a90a I made this backwards compatible by adding the `TableWithSchema` constructor as you suggested.

tomjaguarpaw:
Renzo, I'm afraid super stressed and busy at the moment.  I'm just going to throw out some comments to you.  I'm sorry they're not more refined.

* Instead of `BaseTable String String ...` we should have `BaseTable TableIdentifier ...`
* `data TableIdentifier = TableIdentifier { tiSchema :: Maybe String, tiName :: String }` or something.
* Something similar needs to happen for `SqlTable`, with something like `SqlTableIdentifier`
* `TableIdentifier` should probably be an argument to `defaultSqlUpdate/Insert/Delete`
* Can you get rid of mentions of "public" from the tutorials?  I want to keep them simple for now.
* Please return the test `ConnectInfo` to what it was before.  Yes, I know it's silly, but I'm too stressed to think about changing it for now!
* Your changes to `src/Opaleye/Internal/Table.hs` are very sensible.  Please make them in a separate, self-contained commit (without adding `TableWithSchema` or anything else) and I will merge them separately.

k0001:
Tom, thanks for your feedback.

I'll work on all this, but I'm afraid it will be in a couple of days
because my computer hardware broke and, well, I need access to a working
computer first.

In the meantime, I wish you good health and that your stress goes away :)
Il 25/ago/2015 12:42 PM, "tomjaguarpaw" <notifications@github.com> ha
scritto:

tomjaguarpaw:
Sure, that's fine.  Thanks for the good wishes and good luck with the computer too.

k0001:
I think I've covered all your comments. I've separated the changes in two commits.

tomjaguarpaw:
This looks good.  I will have to take a closer look when I get more  time.

NB There are merge conflicts

k0001:
Thanks. I just rebased the changes against `master`.

tomjaguarpaw:
Excellent stuff!  Thanks.

PR #103, opened 2015-08-22
Render SQL statements in a single line.

k0001:
This change makes it easier to grep for queries in PostgreSQL's log, as well as copy-pasting said queries.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -200,7 +200,7 @@ table4columndata :: [(Column O.PGInt4, Column O.PGInt4)]
 table4columndata = table4dataG
 
 table6data :: [(String, String)]
-table6data = [("xy", "a"), ("z", "a"), ("more text", "a")]
+table6data = [("xy", "a"), ("z", "a"), ("more \n text", "a")]
 
 table6columndata :: [(Column O.PGText, Column O.PGText)]
 table6columndata = map (\(column1, column2) -> (O.pgString column1, O.pgString column2)) table6data
--- a/src/Opaleye/Internal/Print.hs
+++ b/src/Opaleye/Internal/Print.hs
@@ -12,9 +12,15 @@ import           Opaleye.Internal.Sql (Select(SelectFrom, Table,
 import qualified Opaleye.Internal.HaskellDB.Sql as HSql
 import qualified Opaleye.Internal.HaskellDB.Sql.Print as HPrint
 
+import qualified Text.PrettyPrint.HughesPJ as PP 
 import           Text.PrettyPrint.HughesPJ (Doc, ($$), (<+>), text, empty,
                                             parens)
+
 import qualified Data.List.NonEmpty as NEL
+
+
+renderDoc :: Doc -> String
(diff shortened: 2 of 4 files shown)


Test/Test.hs
```diff
@@ -200,7 +200,7 @@ table4columndata :: [(Column O.PGInt4, Column O.PGInt4)]
 table4columndata = table4dataG
 
 table6data :: [(String, String)]
-table6data = [("xy", "a"), ("z", "a"), ("more text", "a")]
+table6data = [("xy", "a"), ("z", "a"), ("more \n text", "a")]
```
k0001: Minor test to check whether newlines are preserved in strings. They do.

tomjaguarpaw:
I'm not opposed to this but I'd prefer to hear from others what they think first.  On the other hand I would happily merge a flag to choose SQL formatting style.

k0001:
Sure, let's discuss it. A couple of more comments/opinion about this:

* This is how the SQL end up looking like in the system log: https://gist.githubusercontent.com/k0001/df57c255c5ad85c3f839/raw/1efa89e65f3c6fcb46e09bed9bed242e90b731be/gistfile1.txt (SQL from the test suite)

* Each line in the SQL is preceded with appropriate indentation whitespace, even if the log above doesn't show it. That whitespace can get large, and it needs to be sent across the network and afterwards parsed and discarded.

* Adding a flag to choose the formatting means we probably need to add new variants of the `runXxx` functions as we don't have a practical way of providing a “global configuration” for these things (which is probably OK). I'd rather have this compact one-line format be the default when using `runQuery` and friends, but have the functions on `Opaleye.Sql` render the pretty-printed version useful for debugging.

WraithM:
It would be great if there were a flag of some sort, like Tom mentioned. Exposing a higher-order function might be a good way to implement this. I like the one-line format should be the default for postgresql consumption, like @k0001 said. I definitely use the pretty printing functions for debugging, and I really like having the indentation for those purposes.

bergmark:
I could see the ability to choose being useful for us. We'd just stick this in our [Config](https://github.com/silkapp/silk-opaleye/blob/master/src/Silk/Opaleye/Config.hs#L30) so our library's interface wouldn't change regardless of opaleye's implementation.

tomjaguarpaw:
I've been super busy, so got behind with PRs.

Renzo, ultimately I like your idea that what is sent over the wire should be on a single line.  We should have two separate functions `formatAndShowSQLNested` and `formatAndShowSQLSingleLine` though, I think, so that people can choose between them.

tomjaguarpaw:
Sorry this has been languishing.  I've remained super busy.

I'm happy for the SQL that goes over the wire to be on a single line.  I think we should preseve the existing behaviour of functions in `Opaleye.Sql`, but add `formatAndShowSQLSingleLine` and `showSqlForPostgresSingleLineExplicit` to be used by `RunQuery`.

Could you make those changes and rebase to HEAD?  Then I'll be happy to merge.

PR #104, opened 2015-08-29
Handle returning zero columns in runInsertReturning and runUpdateReturning

tstat:


tomjaguarpaw:
Good idea!  Can you please put `prepareRowParser` in `Internal.RunQuery` and `ensureReturnColumns` in an internal module too? Ideally we'd merge the latter with `ensureColumns` but that's probably too much work for to little gain right now.

tstat:
Sure thing, I moved `prepareRowParser` to `Internal.RunQuery` and eliminated `ensureReturnColumns` in favor of creating a more general version of `ensureColumns` called `ensureColumnsGen`.

tomjaguarpaw:
That's great.

PR #105, opened 2015-08-30
Slight fix to grammar - which vs that

GetContented:
"x is a y that provides" is better than "x is a y which provides" here, because we're talking about identity. "Which" is a pronoun & determiner used to refer to a previous clause, but "that" is a pronoun used to indroduce a defining or qualifying/restricting clause, in particular one used for talking about identity. Which is used when the clause provides inessential, incidental information. (Not the case here).


tomjaguarpaw:
OK, thanks!

PR #108, opened 2015-09-11
Use absolute locations for tutorial links

danse:
Without this, links will look broken [on Hackage](https://hackage.haskell.org/package/opaleye)


tomjaguarpaw:
Good point.  Thanks!

PR #112, opened 2015-09-19
bump upper bound for postgresql-simple

k0001:
It compiles and the test suite runs just fine.


tomjaguarpaw:
Nice, you are a hero.

PR #119, opened 2015-10-08
Added union

andrewthad:
If this gets merged as is, I'll add in all the other binary operations using the same approach.


tomjaguarpaw:
You need to export the explict versions too (though ideally the unexported stuff should be in the `Internal` module).

andrewthad:
Ah. I'll move that to an internal module.

andrewthad:
It has now been moved and the export list deleted.

tomjaguarpaw:
That's great, thanks.  We've been missing that for a long time!

PR #126, opened 2015-10-29
Adding IN and NOT IN subquery operators.

WraithM:
This adds the `IN` and `NOT IN` operators for subqueries rather than simply `Foldable` haskell data structures as done in issue #21. The new functions introduced here are `columnIn` and `columnNotIn` respectively.

```haskell
columnIn :: Query (Column a) -> QueryArr (Column a) ()
columnNotIn :: Query (Column a) -> QueryArr (Column a) ()
```

Here's an example:

```sql
CREATE TABLE person (
    id serial primary key,
    name text,
    age int4
);

CREATE TABLE test_people (
    person_id int4 references person(id)
);
```

can be queried with:

```haskell
data PersonP i n a = PersonP
    { personId :: i
    , pname    :: n
    , page     :: a
    } deriving (Show, Eq)

makeAdaptorAndInstance "pPerson" ''PersonP

type WritePerson = PersonP
    (Maybe (Column PGInt4))
    (Column PGText)
    (Column PGInt4)

type ReadPerson = PersonP
    (Column PGInt4)
    (Column PGText)
    (Column PGInt4)

personTable :: Table WritePerson ReadPerson
personTable = Table "person"
    (pPerson PersonP
        { personId = optional "id"
        , pname = required "name"
        , page = required "age"
        })

testPeopleTable :: Table (Column PGInt4) (Column PGInt4)
testPeopleTable = Table "test_people" (p1 (required "person_id"))

testQuery :: Query ReadPerson
testQuery = proc () -> do
    row@(PersonP pid _ _) <- queryTable personTable -< ()
    columnNotIn (queryTable testPeopleTable) -< pid
    returnA -< row
```

which produces the following SQL:

```sql
SELECT "id0_1" as "result1_3",
       "name1_1" as "result2_3",
       "age2_1" as "result3_3"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   "name" as "name1_1",
                   "age" as "age2_1"
            FROM "person" as "T1") as "T1",
           (SELECT "person_id" as "person_id0_2"
            FROM "test_people" as "T1") as "T2"
      WHERE (("id0_1") NOT IN ("person_id0_2"))) as "T1"
```

I'm not absolutely certain about the `as "Tx"` part of the code. Please let me know if there is a mistake, and I'd be happy to fix it. However, I tested these queries, and it works as intended.

### DIFF
--- a/src/Opaleye/Internal/HaskellDB/PrimQuery.hs
+++ b/src/Opaleye/Internal/HaskellDB/PrimQuery.hs
@@ -46,7 +46,7 @@ data Literal = NullLit
 
 data BinOp      = OpEq | OpLt | OpLtEq | OpGt | OpGtEq | OpNotEq
                 | OpAnd | OpOr
-                | OpLike | OpIn
+                | OpLike | OpIn | OpNotIn
                 | OpOther String
 
                 | OpCat
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -145,6 +145,7 @@ showBinOp  OpAnd        = "AND"
 showBinOp  OpOr         = "OR"
 showBinOp  OpLike       = "LIKE"
 showBinOp  OpIn         = "IN"
+showBinOp  OpNotIn      = "NOT IN"
 showBinOp  (OpOther s)  = s
 showBinOp  OpCat        = "||"
 showBinOp  OpPlus       = "+"
@@ -157,7 +158,7 @@ showBinOp  OpBitAnd     = "&"
 showBinOp  OpBitOr      = "|"
 showBinOp  OpBitXor     = "^"
 showBinOp  OpAsg        = "="
-
+ 
 
 data UnOpType = UnOpFun | UnOpPrefix | UnOpPostfix
 
--- a/src/Opaleye/Operators.hs
+++ b/src/Opaleye/Operators.hs
@@ -31,6 +31,22 @@ restrict :: QueryArr (Column T.PGBool) ()
 restrict = QueryArr f where
   f (Column predicate, primQ, t0) = ((), PQ.restrict predicate primQ, t0)
 
+{-| Restrict a column using the IN operator. -}
+columnIn :: QueryArr () (Column a) -> QueryArr (Column a) ()
+columnIn (QueryArr q) = QueryArr f
+  where 
+    f (Column colName, primQ', t0) = 
+        let (Column colResult, primQ, t1) = q ((), primQ', t0) in
+        ((), PQ.Product (return primQ) [HPQ.BinExpr HPQ.OpIn colName colResult], t1)
+
+{-| Restrict a column using the NOT IN operator. -}
+columnNotIn :: QueryArr () (Column a) -> QueryArr (Column a) ()
+columnNotIn (QueryArr q) = QueryArr f
+  where 
+    f (Column colName, primQ', t0) = 
+        let (Column colResult, primQ, t1) = q ((), primQ', t0) in
+        ((), PQ.Product (return primQ) [HPQ.BinExpr HPQ.OpNotIn colName colResult], t1)
+
 {-| Filter a 'QueryArr' to only those rows where the given condition
 holds.  This is the 'QueryArr' equivalent of 'Prelude.filter' from the
 'Prelude'.  You would typically use 'keepWhen' if you want to use a
--- a/src/Opaleye/Operators.hs
+++ b/src/Opaleye/Operators.hs
@@ -31,22 +31,6 @@ restrict :: QueryArr (Column T.PGBool) ()
 restrict = QueryArr f where
   f (Column predicate, primQ, t0) = ((), PQ.restrict predicate primQ, t0)
 
-{-| Restrict a column using the IN operator. -}
-columnIn :: QueryArr () (Column a) -> QueryArr (Column a) ()
-columnIn (QueryArr q) = QueryArr f
-  where 
-    f (Column colName, primQ', t0) = 
-        let (Column colResult, primQ, t1) = q ((), primQ', t0) in
-        ((), PQ.Product (return primQ) [HPQ.BinExpr HPQ.OpIn colName colResult], t1)
-
-{-| Restrict a column using the NOT IN operator. -}
-columnNotIn :: QueryArr () (Column a) -> QueryArr (Column a) ()
-columnNotIn (QueryArr q) = QueryArr f
-  where 
-    f (Column colName, primQ', t0) = 
-        let (Column colResult, primQ, t1) = q ((), primQ', t0) in
-        ((), PQ.Product (return primQ) [HPQ.BinExpr HPQ.OpNotIn colName colResult], t1)
-
 {-| Filter a 'QueryArr' to only those rows where the given condition
 holds.  This is the 'QueryArr' equivalent of 'Prelude.filter' from the
 'Prelude'.  You would typically use 'keepWhen' if you want to use a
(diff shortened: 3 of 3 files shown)


tomjaguarpaw:
I think you want at least

    columnIn :: QueryArr () (Column a) -> QueryArr (Column a) (Column Bool)

There's no reason the operation should do restriction too.  It can be generalised further, but that would be a good place to start.

WraithM:
Yeah, that's an interesting point. I just couldn't really figure out a way to write the code that made sense for that.

The reason why I wanted to do it this way is that I believe the `IN` and `NOT IN` operators are only valid in a `WHERE` clause. I don't think that you can make a statement like `SELECT person.id IN test_person.id FROM ...` or anything like that. Basically, I don't think that the `IN` operator produces a boolean value in SQL.

Just to be clear, you're suggesting that using the query to restrict the output should be this:
```haskell
restrict -< columnIn (queryTable testPersonTable) -< pid
```
or something like that?

bergmark:
```
=# select baz in (0,1,2,3) from bar; 
?column?
----------
 t
(1 row)
````

**Comment by WraithM** [comment]
I was wrong! Awesome, thank you.

I'll put my thinking cap on and try to come up with a function that does this.

**Comment by WraithM** [comment]
I just turned `IN` and `NOT IN` to boolean operators. The following produces exactly the same SQL as above.

```haskell
testQuery :: Query ReadPerson
testQuery = proc () -> do
    row@(PersonP pid _ _) <- queryTable personTable -< ()
    personId <- queryTable testPeopleTable -< ()
    restrict -< pid `columnNotIn` personId
    returnA -< row
```

tomjaguarpaw:
I'm rather puzzled by this.  Is this code really what you want?

         WHERE (("id0_1") NOT IN ("person_id0_2")))

This doesn't even look like meaningful SQL.

tomjaguarpaw:
[On the other thread](https://github.com/tomjaguarpaw/haskell-opaleye/issues/21#issuecomment-151955535) you said you wanted

    WHERE column IN (SELECT column FROM tbl_name);
    WHERE column NOT IN (SELECT column FROM tbl_name);

tomjaguarpaw:
I don't think your query means what you think it does.  Here's an SQL strangeness:

    select * from table1 where (300 = "column2");

Does what you think

    select * from table1 where (300 = ("column2"));

Means the same as above.  Parens are for (trivial) precedence.

    select * from table1 where (300 in ("column2"));

`("column2")` is now treated as a sequence ...

    select * from table1 where (300 in "column2");  -- ERROR:  syntax error at or near ""column2""

... and you can't remove the parens.

SQL is really quite annoying.

WraithM:
I did some more tests with this PR, and you're totally right. I thought I was doing the correct query, but upon further inspection, I found that my tests weren't sound for what I actually wanted.

Let me explain my use case, and maybe that can help guide us towards the right solution.

Let's say I have the `person` and `test_people` tables above (my actual use case is more complicated, but let's simplify it to this). I want all `person`s except those where the `id` exist in `test_people`.

One way to write this would be to use the new `except` combinator. eg.
```haskell
exceptQuery :: Query ReadPerson
exceptQuery = 
        queryTable personTable
    `except`
        proc () -> do
            row@(PersonP pid _ _) <- queryTable personTable -< ()
            personId <- queryTable testPeopleTable -< ()
            restrict -< pid .== personId
            returnA -< row
```
The SQL output is:
```sql
SELECT "binary0_4" as "result1_5",
       "binary1_4" as "result2_5",
       "binary2_4" as "result3_5"
FROM (SELECT *
      FROM (SELECT "id0_1" as "binary0_4",
                   "name1_1" as "binary1_4",
                   "age2_1" as "binary2_4"
            FROM (SELECT *
                  FROM (SELECT "id" as "id0_1",
                               "name" as "name1_1",
                               "age" as "age2_1"
                        FROM "person" as "T1") as "T1") as "T1"
            EXCEPT
            SELECT "id0_2" as "binary0_4",
                   "name1_2" as "binary1_4",
                   "age2_2" as "binary2_4"
            FROM (SELECT *
                  FROM (SELECT "id" as "id0_2",
                               "name" as "name1_2",
                               "age" as "age2_2"
                        FROM "person" as "T1") as "T1",
                       (SELECT "person_id" as "person_id0_3"
                        FROM "test_people" as "T1") as "T2"
                  WHERE (("id0_2") = ("person_id0_3"))) as "T1") as "T1") as "T1"
```

Another way to write this would be:

```sql
SELECT * 
FROM person
WHERE person.id NOT IN (SELECT person_id FROM test_people)
```

I thought that the `columnNotIn` query above was equivalent to this last one, but it's really not. It's doing a join between `person` and `test_people` that I really didn't intend. Thank you for pointing out my sloppy thinking. The results from my tests were just coincidentally what I expected... Though, I think the `IN` operator should always work as expected, oddly enough. It's actually only `NOT IN` that won't work this way.

Clearly, the binary operator approach above is wrong, and also the original code that I wrote is also clearly wrong.

The types should be:
```haskell
columnIn :: Query (Column a) -> QueryArr (Column a) (Column PGBool)
columnNotIn :: Query (Column a) -> QueryArr (Column a) (Column PGBool)
```

I'm having trouble creating a `QueryArr` that embeds the query on the right-hand side of the `IN`/`NOT IN` operator. Do you have suggestions there? What `PrimExpr` should I be looking to construct?

WraithM:
I'm a bad person, and I'm scared... I had the random idea to just jam the subquery into a string literal, and uh... it worked...

```haskell
columnNotIn :: QueryArr () (Column a) -> QueryArr (Column a) (Column T.PGBool)
columnNotIn g = QueryArr f
  where
    (Column c, primQ, t) = runSimpleQueryArrStart g ()
    subquery = toHPQ ([c], optimize primQ, t)
    -- Backdoor as string
    toHPQ :: ([HPQ.PrimQuery], PQ.PrimQuery, Tag) -> HPQ.PrimQuery
    toHPQ = HPQ.ConstExpr . HPQ.OtherLit . parens . formatAndShowSQL
    parens x = "(" ++ x ++ ")"
    f :: (Column a, PQ.PrimQuery, Tag) -> (Column T.PGBool, PQ.PrimQuery, Tag)
    f (Column inputCol, primQ', t0) = (Column (HPQ.BinExpr HPQ.OpNotIn inputCol subquery), primQ', t0)
```
used like such:
```haskell
testQuery :: Query ReadPerson
testQuery = proc () -> do
    row@(PersonP pid _ _) <- queryTable personTable -< ()
    isNotInTestPeople <- columnNotIn (queryTable testPeopleTable) -< pid
    restrict -< isNotInTestPeople
    returnA -< row
```
generates:
```sql
       "name1_1" as "result2_2",
       "age2_1" as "result3_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   "name" as "name1_1",
                   "age" as "age2_1"
            FROM "person" as "T1") as "T1"
      WHERE (("id0_1") NOT IN (SELECT "person_id0_1" as "result1_2"
      FROM (SELECT *
          FROM (SELECT "person_id" as "person_id0_1"
                FROM "test_people" as "T1") as "T1") as "T1"))) as "T1"
```
which worked as I intended...

Though, in all seriousness, it's really difficult to get a `Opaleye.Internal.HaskellDB.PrimQuery` from an `Opaleye.Internal.PrimQuery` without doing something like this.

tomjaguarpaw:
You *are* a bad person but it's not your fault, because `PrimQuery` needs a new constructor to support this.

This looks nice now.  It would be good to generalize it to

    columnNotIn :: Something a => QueryArr () a -> QueryArr a (Column T.PGBool)

WraithM:
Wait, which `PrimQuery`? `HPQ.PrimQuery`?

tomjaguarpaw:
There is only one `PrimQuery` type.  The `HaskellDB.PrimQuery` module is called that for historical reasons, but it doesn't contain a `PrimQuery` type anymore.  I'm not sure how

        toHPQ :: ([HPQ.PrimQuery], PQ.PrimQuery, Tag) -> HPQ.PrimQuery

type checked, because.`HPQ.PrimQuery` doesn't exist.  It's `HPQ.PrimExpr`.

WraithM:
Oops! I added that type annotation manually in my comment above to try to make it clear, but I failed.

WraithM:
I edited my comment above.

WraithM:
So, I'm curious what you mean by `PrimQuery` needing a new constructor. Doesn't `PrimExpr` actually need a new constructor? If we're going to make a `Column PGBool` out of a `PrimQuery`, then we need to be able to insert `PrimQuery`s into a `PrimExpr`, no? I don't think I understand what you mean.

tomjaguarpaw:
Hmm, well at least one of them needs a new constructor ...

tomjaguarpaw:
Instead of messing around with the internals I just implemented this is terms of left join: https://github.com/tomjaguarpaw/haskell-opaleye/commit/1df1ad3fdd40b4e19850badc870918217338d340

I'll leave it up to you whether you close this PR or keep it open to make it generate nicer code.  My implementation should give exactly equivalent results though.

WraithM:
I like it! Thanks. Maybe I'll revisit this at some point in the future, but this solution works for now.

Issue #107, opened 2015-09-03
Add some instances for types of the form C a => C (Column a)


djeik:
For instance, if we have `C = IsString`, then we can use the OverloadedStrings extensions to insert `String`, `Text`, and `ByteString` values in user-defined product types, rather than have to use `pgString` . 
There is already an instance like this for Num, which is what lets the tutorial look like it can directly insert integers, whereas in reality the Num instance for columns is doing some work that's not immediately clear in the tutorial.

bergmark:
Also see #80 #81

tomjaguarpaw:
`IsString (Column PGText)` certainly makes sense.  I don't see what you mean "we can use the OverloadedStrings extensions to insert `String`, `Text`, and `ByteString` values in user-defined product types" though.  `IsString` will only help for literals.

djeik:
@tomjaguarpaw I meant that we would be able to use literals to insert those types of values. My wording was pretty unclear though; sorry.

tomjaguarpaw:
Well, technically it wouldn't be a literal of those types, it would be a literal of type `Column (PGText)`.  But yes, this is a good idea.  Feel free to add it next to the `Num` instance.

tomjaguarpaw:
@mitchellwrosen has started off this work on #129.  Feel free to file a new issue if you want to discuss support for further types.

PR #129, opened 2015-12-03
Add IsString (Column ...) instances

mitchellwrosen:
This adds IsString instances for
- Column PGText
- Column PGBytea
- Column PGJson
- Column PGJsonb

Addresses #107

### DIFF
--- a/src/Opaleye/Internal/Column.hs
+++ b/src/Opaleye/Internal/Column.hs
@@ -1,5 +1,7 @@
 module Opaleye.Internal.Column where
 
+import Data.String
+
 import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 
 -- | Numeric 'Column' types are instances of 'Num', so you can use
@@ -64,3 +66,9 @@ class PGFractional a where
 instance (PGNum a, PGFractional a) => Fractional (Column a) where
   fromRational = pgFromRational
   (/) = binOp HPQ.OpDiv
+
+class PGString a where
+    pgFromString :: String -> Column a
+
+instance PGString a => IsString (Column a) where
+  fromString = pgFromString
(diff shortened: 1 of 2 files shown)


bergmark:
Are the json instances partial? I'd like to avoid that.

mitchellwrosen:
I'm think so... they're as partial as `pgJSON` and `pgJSONB`.

tomjaguarpaw:
Thanks for this.  Can you explain why you introduced `PGStringRep` rather than just implementing the `IsString` instances directly?

I'm certainly happy for `PGText` to have an `IsString` instance, but like @bergmark I'd prefer to avoid them being partial, even at the PG-runtime level.  I'm also not entirely conivnced of the utility of the instance for `bytea`, but feel free to persuade me!

mitchellwrosen:
`instance String (Column PGText)` is definitely ideal. I just threw this patch together after wrestling with some cyclic dependency issues. 

Without some restructuring of the modules, I'm not sure what file this instance is supposed to be written in :)

tomjaguarpaw:
There are `Num` instances in `Internal.Column`, so I'd put an `IsString` instance there too.

tomjaguarpaw:
Oh wait, now I understand what you mean about the cyclic dependencies!  OK, this needs some thinking about ...

tomjaguarpaw:
My suggestion is that we keep basically what you've already got, but with

    pgFromString :: String -> Column a

and get rid of the type families.  Let's only provide an instance for `PGText` for now.  If you make those changes and squash your commits into one I'm happy to merge!  Thanks.

mitchellwrosen:
Do you mean this?

```haskell
class PGString a where
    pgFromString :: String -> Column a

instance PGString a => IsString (Column a) where
    fromString = pgFromString

instance PGString PGText where
    pgFromString = pgString
```

tomjaguarpaw:
Yes, exactly that!

mitchellwrosen:
Sure thing, updated.

tomjaguarpaw:
Cool, thanks.

Issue #151, opened 2016-04-07
Constants fail in conversion from SQL when too polymorphic


hesselink:
If you use a constant in a context where it's not constrained to a type, you get an error when converting it to a Haskell value. For example:

```haskell
{-# LANGUAGE OverloadedStrings #-}

import Data.CaseInsensitive (CI)
import Data.Text (Text)
import Opaleye
import Database.PostgreSQL.Simple (connectPostgreSQL)

test :: IO [CI Text]
test = do
  conn <- connectPostgreSQL ""
  runQuery conn (pure $ constant ("hello" :: CI Text) :: Query (Column PGCitext))
```

This will give:

```
*** Exception: Incompatible {errSQLType = "unknown", errSQLTableOid = Nothing, errSQLField = "result1_1", errHaskellType = "CI Text", errMessage = ""}
```

Sometimes postgres will also default to something instead of giving "unknown". For example, I implemented array literals (would you like me to contribute that back BTW?) and inside array literals, the textual value above will default to 'text' (not 'citext').

It looks to me like the solution would be to always add a type annotation to literals. It could probably be added to the `Constant` class, although that would mean duplication when multiple haskell types map to the same postgres type. In our silk-db-opaleye package we have a type family `PGRep` mapping from Haskell type to postgres type, which we could probably use to remove that duplication.

tomjaguarpaw:
> I implemented array literals (would you like me to contribute that back BTW?)

Yes please!

hesselink:
Cool, I'll open a PR to discuss it further, since there are some rough edges.

tomjaguarpaw:
I agree this is quite annoying.  Adding types to all generated literals is possible, although it will make the generated code somewhat awkward looking.

A short term workaround would be to use `unsafeCast` http://haddock.stackage.org/lts-5.4/opaleye-0.4.2.0/Opaleye-Column.html

hesselink:
Yes, that's what I'm using for now.

hesselink:
Small update since it was mentioned here a few comments up: I sent the array literal generation PR in #154. That includes always casting them with a type class, so it might be relevant to this ticket as well.

Issue #81, opened 2015-07-22
Rename pg..., PG... to sql..., Sql...


tomjaguarpaw:
In anticipation of `opaleye-sqlite` change `pg` and `PG` in names to `sql` and `Sql` instead.  Examples
- `PGTypes` module becomes `SqlTypes`
- `PGText` type becomes `SqlText`
- `pgText` function becomes `SqlText`

tomjaguarpaw:
Migrating to https://github.com/tomjaguarpaw/haskell-opaleye/issues/326

PR #154, opened 2016-04-15
Array literals

hesselink:
This add support for creating arrays in several ways:

* Directly with `pgArray`.
* Using `constant`.
* By manipulating them with `emptyArray`, `singletonArray` and `arrayPrepend`.

This adds an explicit type cast to every array literal, since it's been my experience that without it, many things don't work: empty arrays never work, and array contents tend to default (e.g. citext to text, numbers to int I think). To do this I created a type class `IsPGType` that allows printing the postgres type string from the haskell postgres type.

The one bit that is a little nasty is that `pgArray` takes a function to create the inner literal, and this has to be one of the functions creating a literal, but that's not enforced by the types. This could be fixed by having all those functions return a `Literal`, although it seems some return a cast instead (so with the current code arrays of date/times will not work). What do you think the nicest solution is here?

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -674,12 +674,12 @@ testEmptyArray :: Test
 testEmptyArray = testG (A.pure $ O.pgArray O.pgInt4 []) (== [[] :: [Int]])
 
 -- This test fails without the explicit cast in pgArray since postgres
--- defaults the text to 'text' but postgresql-simple expects 'citext'.
+-- defaults the numbers to 'integer' but postgresql-simple expects 'float8'.
 
-testCITextArray :: Test
-testCITextArray = testG (A.pure $ O.pgArray O.pgCiStrictText texts) (== [texts])
+testFloatArray :: Test
+testFloatArray = testG (A.pure $ O.pgArray O.pgDouble doubles) (== [doubles])
   where
-    texts = ["hello", "world"]
+    doubles = [1 :: Double, 2]
 
 allTests :: [Test]
 allTests = [testSelect, testProduct, testRestrict, testNum, testDiv, testCase,
@@ -694,7 +694,7 @@ allTests = [testSelect, testProduct, testRestrict, testNum, testDiv, testCase,
             testKeywordColNames, testInsertSerial, testInQuery, testAtTimeZone,
             testStringArrayAggregateOrdered, testMultipleAggregateOrdered,
             testOverwriteAggregateOrdered, testCountRows0, testCountRows3,
-            testArrayLiterals, testEmptyArray, testCITextArray
+            testArrayLiterals, testEmptyArray, testFloatArray
             ]
 
 -- Environment.getEnv throws an exception on missing environment variable!
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE Arrows #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 module Main where
 
@@ -662,6 +663,24 @@ testAtTimeZone = testG (A.pure (O.timestamptzAtTimeZone t (O.pgString "CET"))) (
         t' = Time.LocalTime d (Time.TimeOfDay 2 0 0)
         d = Time.fromGregorian 2015 1 1
 
+testArrayLiterals :: Test
+testArrayLiterals = testG (A.pure $ O.pgArray O.pgInt4 vals) (== [vals])
+  where vals = [1,2,3]
+
+-- This test fails without the explicit cast in pgArray since postgres
+-- can't determine the type of the array.
+
+testEmptyArray :: Test
+testEmptyArray = testG (A.pure $ O.pgArray O.pgInt4 []) (== [[] :: [Int]])
+
+-- This test fails without the explicit cast in pgArray since postgres
+-- defaults the text to 'text' but postgresql-simple expects 'citext'.
+
+testCITextArray :: Test
+testCITextArray = testG (A.pure $ O.pgArray O.pgCiStrictText texts) (== [texts])
+  where
+    texts = ["hello", "world"]
+
 allTests :: [Test]
 allTests = [testSelect, testProduct, testRestrict, testNum, testDiv, testCase,
             testDistinct, testAggregate, testAggregate0, testAggregateFunction,
@@ -674,7 +693,8 @@ allTests = [testSelect, testProduct, testRestrict, testNum, testDiv, testCase,
             testValuesEmpty, testUnionAll, testTableFunctor, testUpdate,
             testKeywordColNames, testInsertSerial, testInQuery, testAtTimeZone,
             testStringArrayAggregateOrdered, testMultipleAggregateOrdered,
-            testOverwriteAggregateOrdered, testCountRows0, testCountRows3
+            testOverwriteAggregateOrdered, testCountRows0, testCountRows3,
+            testArrayLiterals, testEmptyArray, testCITextArray
             ]
 
 -- Environment.getEnv throws an exception on missing environment variable!
--- a/src/Opaleye/Column.hs
+++ b/src/Opaleye/Column.hs
@@ -1,12 +1,13 @@
 module Opaleye.Column (module Opaleye.Column,
                        Column,
                        Nullable,
+                       unsafeCast,
                        unsafeCoerce,
                        unsafeCoerceColumn,
                        unsafeCompositeField)  where
 
 import           Opaleye.Internal.Column (Column, Nullable, unsafeCoerce, unsafeCoerceColumn,
-                                          unsafeCompositeField)
+                                          unsafeCast, unsafeCompositeField)
 import qualified Opaleye.Internal.Column as C
 import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 import qualified Opaleye.PGTypes as T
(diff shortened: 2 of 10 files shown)


tomjaguarpaw:
This looks great, thanks.

> The one bit that is a little nasty is that pgArray takes a function to create the inner literal, and this has to be one of the functions creating a literal, but that's not enforced by the types.

Why does it have to create a `Literal`?  Wouldn't it be OK to use `defaultSqlExpr` rather than `defaultSqlLiteral`?

The other thing I want to think hard about is the consequences of this `IsPGType` typeclass.  I'd like to avoid it if at all possible.  I will see if I think we can.

tomjaguarpaw:
Questions and issues to address before merging:

1. Can you explain why `ScopedTypeVariables` is needed in `Constant.hs`?

2. Why does it have to create a `Literal`? Wouldn't it be OK to use `defaultSqlExpr` rather than `defaultSqlLiteral`?

3. I think I would prefer all the `pgString`, `pgInt4` functions to contain casts rather than introduce the `IsPGType` typeclass, i.e. every creation of a literal contains its type.  What do you think?

tomjaguarpaw:
Also, it needs some tests.

hesselink:
> Can you explain why ScopedTypeVariables is needed in Constant.hs?

I was previously, but then I found `constantExplicit`, and now it's not. I've removed it.

> Why does it have to create a Literal? Wouldn't it be OK to use defaultSqlExpr rather than defaultSqlLiteral?

I looked at `defaultSqlExpr` but it returned an `SqlExpr`, not a `String`, so I dismissed it. But I guess I can use `ppSqlExpr` and then convert the `Doc` to a `String`?

> I think I would prefer all the pgString, pgInt4 functions to contain casts rather than introduce the IsPGType typeclass, i.e. every creation of a literal contains its type. What do you think?

I thought you previously didn't like that idea (in #151) since it would make the generated sql too noisy. I've also come to agree with that: it would be annoying to see something like `WHERE x = (3 :: int)`, although not horrible. With casts on the literals, we'd still also need them on empty arrays, though, which means we'd still need a way to go from Haskell type to postgres type string.

What is your objection to the type class?

tomjaguarpaw:
> I was previously, but then I found constantExplicit, and now it's not. I've removed it.

Cool, thanks.

> I've removed it.

Have you pushed it?  I can't see a change.  Perhaps you need to reissue the PR?

> I guess I can use ppSqlExpr and then convert the Doc to a String?

On second thoughts we really need an array literal constructor in `PrimExpr`.

> make the generated sql too noisy

I agree it's more noisy and not desirable, all other things being equal, but making the surface syntax noisy seems even worse.  Theoretically we could pass over the AST before generating the SQL to remove all redundant casts.  Thoughts?

hesselink:
> Have you pushed it? I can't see a change. Perhaps you need to reissue the PR?

I hadn't then, I've pushed now and the PR has been updated.

> > I guess I can use ppSqlExpr and then convert the Doc to a String?
> On second thoughts we really need an array literal constructor in PrimExpr.

Ah, that's a good idea. So it wouldn't be a "literal" per se, since it could include e.g. references to columns, but that's fine. I'll do that.

> I agree it's more noisy and not desirable, all other things being equal, but making the surface syntax noisy seems even worse. Theoretically we could pass over the AST before generating the SQL to remove all redundant casts. Thoughts?

I still think it's moot until there's a solution for empty arrays. The only ones I see are the current one, and passing the type as a string when constructing empty arrays. However, since they're used in many places, that string argument would show up wherever the `IsPGType` class now shows up, I think.

tomjaguarpaw:
So what do we need to do before merging?

1. Add an array literal constructor to `PrimExpr`.  I think this should be straightforward.  It will probably end up looking a lot like the existing `ListExpr` constructor, though with `[` instead of `(`.

2. Add tests

3. Decide what we'll do about this annoying types issue.  I agree that the empty array case still presents problems.

hesselink:
Yes, sounds right. I'm working on 1 now. I'll do 2 after, and we can think about 3 in the meantime.

hesselink:
OK, 1 is done and looks pretty clean. I have to stop now, I'll work on the tests next time I have some free time.

tomjaguarpaw:
That looks great, thanks.

hesselink:
I've added some tests, one just creating a simple array literal, the other two testing failure cases I found while using this: empty arrays lacking a type, and arrays of 'citext' defaulting to 'text', and failing on deserialization in Haskell.

tomjaguarpaw:
Great!  OK, so what do we do about the types issue?  Your observation about empty arrays makes me think that perhaps `IsPGType` is the right way to go.

hesselink:
I think it is, but I never clearly heard what your objection to the type class was. Is it just the noise of having the constraint on the array functions?

tomjaguarpaw:
I don't mind the noise so much, but I wanted to avoid adding another typeclass if at all possible.  It seems that it's not possible though and it's actually quite a natural typeclass.

hesselink:
I agree. In that case, unless you think there's something missing, I guess this is done.

tomjaguarpaw:
Minor point:  Why is `pgArray`

    pgArray :: forall a b. IsPGType b => (a -> C.Column b) -> [a] -> C.Column (PGArray b)

rather than

    pgArray :: forall a b. IsPGType b => [C.Column b] -> C.Column (PGArray b)

I don't particularly mind the former, but the latter seems more natural to me.

hesselink:
That's a good question... I guess the former felt more natural to me :) The functions in `PGTypes` seem to take a haskell value to a postgres value. That is why it takes a list of things. Then it has to take a function for the elements. This is very similar to how a type class would do it. It also makes the implementation of `Default Constant [a] (PGArray b)` very clean. But I wouldn't mind very much if you want to change it.

tomjaguarpaw:
I've had a cunning idea.

The problem with typing `[]` is similar to the problem of creating an empty subselect.  It's hard to make a generic empty subselect because SQL doesn't know the type of its columns. Opaleye *does* allow you to represent [empty subselects in its `PrimQuery`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/PrimQuery.hs#L39) but we [remove them in `Optimize`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/Optimize.hs#L35) before generating SQL.  This is possible because we know all possible results of combining `Empty` with any other query.  We could do the same thing for empty arrays and therefore avoid adding `IsPGType`!

hesselink:
How would you remove something like `SELECT array_prepend(1, ARRAY[]) ...`?

Additionally, there's the issue of defaulting in non-empty arrays as well, although this might also apply to other constants. E.g. if you have a constant array of `citext`, postgres will return it with type `text[]` instead, and deserialization in postgresql-simple will fail because the postgres type doesn't match the Haskell type.

tomjaguarpaw:
* `array_prepend(1, ARRAY[])` would be rewritten as `[1]`.

* We can solve the problem of defaulting in non-empty arrays with the (messy) approach of giving types to every constant (and potentially removing the types that we know are redundant).  I would rather the surface syntax be clean than the generated code.

* It is harder to remove `SELECT []` but I'm confident it can be done.

hesselink:
OK, but now the optimizer needs knowledge of `array_prepend` and potentially other (even user defined) functions to make sure it removes all empty arrays. Otherwise there will be errors lurking in all non-common use cases.

Note that the surface syntax is only "dirty" (debatable) for the handful of array functions. And like you said the class itself is pretty obvious. This kind of optimization, on the other hand, will result in very dirty generated SQL, as well as a large, non-obvious optimization step that will make the internals "dirty", decreasing maintainability and approachability. I'm not convinced it's the right trade-off.

tomjaguarpaw:
Yes, knowledge of potential user defined functions is one place where this falls down.  I'll give it some more thought.

tomjaguarpaw:
OK, I'm convinced of the need for `IsPGType` now.  I'm going to need it to fix https://github.com/tomjaguarpaw/haskell-opaleye/issues/163

Please call it `IsSqlType` though.  I'm going to replace mentions of "PG" with "Sql" https://github.com/tomjaguarpaw/haskell-opaleye/issues/81

hesselink:
Done.

tomjaguarpaw:
Needs rebasing?

tomjaguarpaw:
Happy to merge when that's done.

hesselink:
Conflict with my other PR :) Fixed now.

tomjaguarpaw:
Test failure: `type \"citext[]\" does not exist`

Any ideas?

hesselink:
Yes, I understand why that happens. The `citext` type doesn't exist by default, it's an extension you have to enable with `CREATE EXTENSION citext`. I can see several options:

* Create the extension when setting up the test database.
* Find a standard type that reproduces the same behavior.
* Drop the test case.

I'm leaning towards the first. What do you think?

tomjaguarpaw:
I lean towards the first too, but if you know a standard type that reproduces the same behaviour then go ahead and use it.  I'm happy for you to just enable the extension in the tests though, if you want.

hesselink:
I don't know a standard type that reproduces this, so that's why I was leaning towards creating the extension. However, I found that you need to be 'superuser' to do this. Do you mind if the tests are only runnable as a superuser?

tomjaguarpaw:
Oh, that's getting somewhat complicated.  Will the `int`/`float` distinction work as a test for this?

hesselink:
I'll give it a try.

hesselink:
Good call, that worked. The tests should now no longer depend on the `citext` type.

tomjaguarpaw:
Nice.  Good to have array literals in Opaleye.

PR #157, opened 2016-04-25
Add default method signature to QueryRunnerColumnDefault

cdepillabout:
Add a default method signature to `QueryRunnerColumnDefault` that uses
postgresql-simple's `FromField` class:

```haskell
class QueryRunnerColumnDefault pgType haskellType where
  queryRunnerColumnDefault :: QueryRunnerColumn pgType haskellType
  default queryRunnerColumnDefault :: FromField haskellType => QueryRunnerColumn pgType haskellType
  queryRunnerColumnDefault = fieldQueryRunnerColumn
```
This makes it slightly simpler to define QueryRunnerColumnDefault
instances for types that already have instances for `FromField`:

```haskell
instance QueryRunnerColumnDefault T.PGInt8 Int64 where
  queryRunnerColumnDefault = fieldQueryRunnerColumn

-- becomes

instance QueryRunnerColumnDefault T.PGInt4 Int32
```

### DIFF
--- a/src/Opaleye/Internal/RunQuery.hs
+++ b/src/Opaleye/Internal/RunQuery.hs
@@ -119,6 +119,24 @@ instance QueryRunnerColumnDefault a b =>
 -- | A 'QueryRunnerColumnDefault' @pgType@ @haskellType@ represents
 -- the default way to turn a @pgType@ result from the database into a
 -- Haskell value of type @haskelType@.
+--
+-- Adding a 'QueryRunnerColumnDefault' instance for your own type can be as
+-- simple as defining an instance of 'FromField' for your type.
+--
+-- Let's assume we have a type called Foo:
+--
+-- > newtype Foo = Foo String
+--
+-- We can create an instance of 'QueryRunnerColumnDefault' for Foo by using
+-- 'QueryRunnerColumnDefault''s default method, as long as we have an instance
+-- of 'FromField' for Foo.
+--
+-- > instance FromField Foo where
+-- >   fromField :: FieldParser Foo
+-- >   fromField field Nothing = returnError UnexpectedNull field ""
+-- >   fromField field (Just foo) = pure . Foo $ ByteString.unpack foo
+-- >
+-- > instance QueryRunnerColumnDefault PGText Foo
 class QueryRunnerColumnDefault pgType haskellType where
   queryRunnerColumnDefault :: QueryRunnerColumn pgType haskellType
   default queryRunnerColumnDefault :: FromField haskellType => QueryRunnerColumn pgType haskellType
--- a/src/Opaleye/Internal/RunQuery.hs
+++ b/src/Opaleye/Internal/RunQuery.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleContexts, FlexibleInstances, MultiParamTypeClasses #-}
+{-# LANGUAGE DefaultSignatures, FlexibleContexts, FlexibleInstances, MultiParamTypeClasses #-}
 
 module Opaleye.Internal.RunQuery where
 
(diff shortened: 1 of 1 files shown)


cdepillabout:
In commit 41bf89c, I add an explanation of how to create an instance of QueryRunnerColumnDefault, making use of the default method signature.


When I tried to create a query using one of my own types, I naively tried to do something like this:

```haskell
data Foo = Foo Text

data Bar' a b c = Bar
    { barId :: a
    , barFoo :: b
    }
    deriving Show
type Bar = Bar' Text Foo
type BarColumn =
    Bar' (Column PGText) (Column PGText) (Column PGText)

$(makeAdaptorAndInstance "pBar" ''Bar')

barTable :: Table BarColumn BarColumn
barTable =
    Table "bar"
        (pBar Bar
            { barId = required "bar_id"
            , barFoo = required "foo"
            })

barQuery :: Query BarColumn
barQuery = queryTable barTable

tryQuery = runQuery barQuery
```

However, the call to `runQuery` would throw an error like this:

```
No instance for (Opaleye.Internal.RunQuery.QueryRunnerColumnDefault
                   Opaleye.PGTypes.PGText MyModule)
  arising from a use of ‘runQuery’
In the expression:
    runQuery barQuery :: IO [Bar]
In an equation for ‘it’:
    it = runQuery barQuery :: IO [Bar]
```

I immediately looked up `QueryRunnerColumnDefault` in the Haddocks, but as it was my first time using Opaleye, I wasn't immediately sure of how to go about creating this instance.

After some digging around, I eventually got it working by doing something like this:

```haskell
instance QueryRunnerColumnDefault PGText Foo where
    queryRunnerColumnDefault :: QueryRunnerColumn PGText Foo
    queryRunnerColumnDefault = fieldQueryRunnerColumn

instance FromField Foo where
    fromField :: FieldParser Foo
    -- fromField :: Field -> Maybe ByteString -> Conversion Foo
    fromField field Nothing = returnError UnexpectedNull field ""
    fromField field (Just foo) = pure . Foo . Text.pack . ByteString.unpack foo
```

I think adding the default method to `QueryRunnerColumnDefault` and the additional documentation will make it a nicer experience for future users.

tomjaguarpaw:
Why not `fromField = fmap (Foo . ByteString.unpack) . fromField`?

tomjaguarpaw:
Hello, thanks for that.

Your default implementation of the typeclass method seems reasonable.  However, I don't really want to over-encourage use of postgresql-simple or the creation of `QueryRunnerColumnDefault`s.

There's already [`queryRunnerColumn`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/RunQuery.hs#L95) for doing what you are trying to do.  The documentation was broken before but I've just fixed it.  In fact it is almost exactly what you are trying to do.  Instead of

    instance QueryRunnerColumnDefault Foo Foo where
       queryRunnerColumnDefault =
           queryRunnerColumn ('Opaleye.Column.unsafeCoerceColumn'
                                  :: Column Foo -> Column PGInt4)
                             Foo
                             queryRunnerColumnDefault

which appears in the documentation you should use

    instance QueryRunnerColumnDefault PGText Foo where
       queryRunnerColumnDefault =
           queryRunnerColumn id Foo queryRunnerColumnDefault

Please let me know how you get on with that.

cdepillabout:
> Your default implementation of the typeclass method seems reasonable. However, I don't really want to over-encourage use of postgresql-simple or the creation of `QueryRunnerColumnDefaults`.

Okay, this sounds good to me.

> There's already `queryRunnerColumn` for doing what you are trying to do. The documentation was broken before but I've just fixed it.

`queryRunnerColumn` looks like it might work, but I guess I should have given a better example of what I'm actually trying to do.  

I'm actually trying to use the [`EmailAddress`](https://hackage.haskell.org/package/email-validate-2.2.0/docs/Text-Email-Validate.html#t:EmailAddress) type from the email-validate package.

I'm also trying to use a smart constructor (`emailAddress`)[https://hackage.haskell.org/package/email-validate-2.2.0/docs/Text-Email-Validate.html#v:emailAddress] to construct the `EmailAddress` type.

Is there some way to use a smart constructor with `queryRunnerColumn`?

Here is the code I currently have using `FromField`:

```haskell
import Database.PostgreSQL.Simple.FromField
    ( Conversion, FieldParser, FromField(..), ResultError(..), returnError )
import Opaleye
    ( PGText, QueryRunnerColumn, QueryRunnerColumnDefault(..)
    , fieldQueryRunnerColumn )
import Text.Email.Validate (EmailAddress, emailAddress)

instance QueryRunnerColumnDefault PGText EmailAddress where
    queryRunnerColumnDefault :: QueryRunnerColumn PGText EmailAddress
    queryRunnerColumnDefault = fieldQueryRunnerColumn

instance FromField EmailAddress where
    fromField :: FieldParser EmailAddress
    -- fromField :: Field -> Maybe ByteString -> Conversion EmailAddress
    fromField field Nothing = returnError UnexpectedNull field ""
    fromField field (Just email) = maybe err pure $ emailAddress email
      where
        err :: Conversion EmailAddress
        err = returnError ConversionFailed field $
            "Could not convert " <> show email <> " to email address"
```

cdepillabout:
Oh yeah, one more thing.  I think it would be nice to add a little note to the QueryRunnerColumnDefault class about using `queryRunnerColumn`.

The GHC error I posted is specifically warning about no `QueryRunnerColumnDefault` instance, so I think that's where most users will immediately look.  It would be nice if there was a hint that they should look at the example in the documentation for `queryRunnerColumn`.

cdepillabout:
One last thing.  It looks like `QueryRunnerColumn` could have a `Functor` instance?

```haskell
instance Functor (QueryRunnerColumn a) where
    fmap = queryRunnerColumn id
```

This might make it a little easier to define `QueryRunnerDefaultColumn` instances for the common case of just changing the output type.

tomjaguarpaw:
A `Functor` instance is definitely a good idea.  https://github.com/tomjaguarpaw/haskell-opaleye/issues/158

tomjaguarpaw:
Are you saying that your current `EmailAddress` code is behaving as you want it, and you're asking if there's a better way to do the Opaleye-related bits?

cdepillabout:
> Are you saying that your current EmailAddress code is behaving as you want it, and you're asking if there's a better way to do the Opaleye-related bits?

Yes, exactly this.

The current `EmailAddress` code is working as I want.  It took a relatively long time to figure it out, so I thought I would send in this pull request to clear up how to do it for future users.

I'd definitely appreciate hearing if there is a more canonical way to do it (for instance, not using `FromField`).

tomjaguarpaw:
OK great.  Making it easier for others in the future is a good goal and I'm fully on board.  How shall we do this?

Your case is more complicated than `queryRunnerColumn` can handle because you want to be able to set the error message.  I've got a few comments.  What do you think about each?

1. You could easily return a `Maybe EmailAddress` using `queryRunnerColumn` and then check for `Just`/`Nothing` later.
2. Instead of writing `instance FromField Foo` and then using `queryRunnerColumnDefault = fieldQueryRunnerColumn` you can write `fooFieldParser :: FieldParser Foo` and then `queryRunnerColumnDefault = fieldParserQueryRunnerColumn fooFieldParser`.  This involves a tiny bit less "typeclass magic" so it seems slightly preferable.
3. When explaining the possible options we could use a three-tiered approach
    1. If you already have a `FromField` instance use `fieldQueryRunnerColumn`
    2. If you have a simple case use `queryRunnerColumn`
    3. If you have a more complicated case but not `FromField` instance write a `FieldParser Foo` and use `fieldParserQueryRunnerColumn`.  You can also add a `FromField` instance using this.

cdepillabout:
Let me respond to each of your 3 points.

> You could easily return a Maybe EmailAddress using queryRunnerColumn and then check for Just/Nothing later.

This is certainly possible.  Depending on your business logic and your application, this may be the best choice in some situations.

For my situation with the `EmailAddress` type, I'm relatively certain that all email addresses in the database will be valid email addresses, so I'd prefer to handle invalid email addresses as something like an IO exception, not as a `Maybe` type.

> Instead of writing instance FromField Foo and then using queryRunnerColumnDefault = fieldQueryRunnerColumn you can write fooFieldParser :: FieldParser Foo and then queryRunnerColumnDefault = fieldParserQueryRunnerColumn fooFieldParser. This involves a tiny bit less "typeclass magic" so it seems slightly preferable.

Yeah, this definitely seems like a good idea. :+1:

> When explaining the possible options we could use a three-tiered approach:
> 1. If you already have a FromField instance use fieldQueryRunnerColumn
> 2. If you have a simple case use queryRunnerColumn
> 3. If you have a more complicated case but not FromField instance write a FieldParser Foo and use fieldParserQueryRunnerColumn. You can also add a FromField instance using this.

Yeah, this sounds good to me.

How do you want to handle updating the documentation?  Do you want to go ahead and add this, and then close this pull request?  Or do you want me to edit this pull request and add this documentation somewhere?


One other thing I was wondering was how to insert my own custom data types.  After looking at the documentation, it looks like I need a `Default Constant` instance for my types.  I ended up writing something like this for `EmailAddress`:

```haskell
import Data.Profunctor (lmap)
import Data.Profunctor.Product.Default (Default(def))
import Database.PostgreSQL.Simple.FromField (Conversion)
import Opaleye (Column, Constant, PGText)
import Text.Email.Validate (EmailAddress, toByteString)

instance Default Constant EmailAddress (Column PGText) where
    def :: Constant EmailAddress (Column PGText)
    def = lmap (decodeUtf8 . toByteString) def
```

I then defined an `insert` method like this:

```haskell
import Data.Profunctor.Product.Default (Default)
import Opaleye ( Constant, PGText, Table, constant, runInsert )

insert
    :: ( Default Constant haskells columns )
    => Connection
    -> Table columns columns'
    -> haskells
    -> IO Int64
insert conn table haskells = runInsert conn table $ constant haskells
```

It might be worthwhile to add something to one of the tutorials about doing a "roundtrip" to and from the database for a custom type (like the `EmailAddress` type above).

As a new user, the tutorials were really helpful, but I don't think they describe how to work with a custom datatype like above.

tomjaguarpaw:
> >  You could easily return a Maybe EmailAddress using queryRunnerColumn and then check for Just/Nothing later.

> This is certainly possible. Depending on your business logic and your application, this may be the best choice in some situations.

tomjaguarpaw:
> >  You could easily return a Maybe EmailAddress using queryRunnerColumn and then check for Just/Nothing later.

> This is certainly possible. Depending on your business logic and your application, this may be the best choice in some situations.

Fair enough.

tomjaguarpaw:
> How do you want to handle updating the documentation? Do you want to go ahead and add this, and then close this pull request? Or do you want me to edit this pull request and add this documentation somewhere?

Either you can update the PR along the lines we discussed or I can do it when I get round to it.  It's totally up to you. It might take me a few weeks though so I'd welcome any help you can give.

Adding an example for creating a `Constant` instance is also a good idea so feel free to add that too.

cdepillabout:
Documentation for `QueryRunnerColumnDefault` has been added in #179 and documentation for `Constant` has been added in #180, so I am closing this PR.

The only thing left to do might be to add a tutorial describing how to take a custom type on a round-trip to and from the database.

PR #165, opened 2016-05-05
Add the ability to distinctily aggregate

### LINKED ISSUE #164: Support for COUNT(DISTINCT x)
I can only see the `count` aggregate, but that doesn't let me apply the distinct condition on the expression being counted.

ocharles:
Fixes #164

### DIFF
--- a/src/Opaleye/Table.hs
+++ b/src/Opaleye/Table.hs
@@ -4,7 +4,7 @@ module Opaleye.Table (module Opaleye.Table,
                       -- * Other
                       View,
                       Writer,
-                      Table(Table, TableWithSchema),
+                      T.Table(T.Table, T.TableWithSchema),
                       TableProperties) where
 
 import           Opaleye.Internal.Column (Column(Column))


tomjaguarpaw:
Cool, thanks!  Hopefully I'll be able to look over this tonight.

ocharles:
I have some code doing max/min/countDistinct in my wrapper over opaleye and it seems to do the right thing

tomjaguarpaw:
Could you add a test in `Test.hs`?  I think you could just adjust/expand one of the existing tests rather than add a whole new one.

tomjaguarpaw:
This looks good.  Some comments:

1. It will need at least one test in `Test.hs`.

2. Please change the definition of `distinct'` to a `case`: https://github.com/tomjaguarpaw/haskell-opaleye/pull/165/files#diff-cc1cf4700a323e9e566917c89269d428R126

3. Will `aggregateDistinct` be sufficient for you?  It will force *all* aggregators to be done with `DISTINCT`.  If this is really all you want why not use `distinct` before doing a normal aggregation?

ocharles:
No, I want to count distinct and take a max and min in the same
aggregation. However I don't think I use the Opaleye surface API (will have
to check) for this so you might be OK changing it.

tomjaguarpaw:
On Thu, May 05, 2016 at 03:37:33PM -0700, Oliver Charles wrote:
> No, I want to count distinct and take a max and min in the same
> aggregation. However I don't think I use the Opaleye surface API (will have
> to check) for this so you might be OK changing it.

I'm not sure what you mean.  Can you be more precise?

ocharles:
About which bit? I am writing opaleye-tf which is mostly using .Internal
modules

tomjaguarpaw:
On Fri, May 06, 2016 at 01:26:50AM -0700, Oliver Charles wrote:
> About which bit? I am writing opaleye-tf which is mostly using .Internal
> modules

Do you mean you're not using the `aggregageDistinct` that you wrote?

ocharles:
Not at the moment, no - we have our own `Aggregator` like thing: https://github.com/ocharles/opaleye-tf/blob/d8f4c7e36573b120afe9b42eccfc531a2e0d2514/src/Opaleye/TF.hs#L585.

ocharles:
I'm not quite sure what you're saying about `aggregateDistinct`. You say it will force all aggregations to be done with distinct, but surely if I just do something like '(,) <$> lmap foo count <*> lmap bar aggregateDistinct count` I would only distinctly count bar? That was my intention, anyway.

tomjaguarpaw:
Sorry, I misunderstood what it does.  Could you call it `distinctAggregator` then, so it doesn't resemble `aggregateOrdered`?

tomjaguarpaw:
That patch failed due to something unrelated.  I've kicked off the build again.

ocharles:
Could this be merged?

ocharles:
Oh bah, I seem to have messed up my branch.

tomjaguarpaw:
?

tomjaguarpaw:
Oh I see.  You've reopened it here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/278

PR #171, opened 2016-05-20
Add charLength function

hesselink:
I chose to reuse the `PGString` class as a constraint to be able to restrict this to `PGText` and `PGCitext. If you would rather have a new type class for this, let me know.


tomjaguarpaw:
Using `PGString` seems good, thanks.

PR #172, opened 2016-05-20
Add quot_ and rem_.

hesselink:
I made these applicable only to integral types, since that mirrors what Haskell does, and the behavior is different: `(5::integer) / 2 == 2`, whereas (5 :: float) / 2 == 2.5. However, the underlying postgres operators work on any numeric type, so let me know if you want me to change that.


tomjaguarpaw:
Just integral types seems fine.

PR #180, opened 2016-06-09
Adding documentation on constant function explaining how to insert custom types into the database.

cdepillabout:
This PR adds some documentation to the `constant` function explaining how to use it to insert custom types into the database as per https://github.com/tomjaguarpaw/haskell-opaleye/pull/157#issuecomment-215823546.

I don't think my explanation in the comment is really that great, so let me know if there is anything that should be changed.


tomjaguarpaw:
Seems like a good start at least.  Feel free to tweak it later.

cdepillabout:
Thanks!

PR #182, opened 2016-06-14
Remove unnecessary code duplication for the PGArray parser.

lostbean:
This commit will avoid the code to get out of sync with postgresql-simple (for instance , there is a up coming [bug fix](https://github.com/lpsmith/postgresql-simple/pull/189))


tomjaguarpaw:
Thanks!

Please make sure it works all the way back to 7.6 (`Applicative` needs to be imported, for example).

tomjaguarpaw:
I did this in https://github.com/tomjaguarpaw/haskell-opaleye/commits/master

PR #183, opened 2016-07-06
Opaleye: generate valid sql with 0 CASE alternatives.

hesselink:
I ran into this when dynamically generating the list of case alternatives, and there happened to be 0 at some point.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -334,6 +334,16 @@ testCase = testG q (== expected)
         expected :: [Int]
         expected = [12, 12, 21, 33]
 
+-- This tests case_ with an empty list of cases, to make sure it generates valid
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Excellent, thanks.

tomjaguarpaw:
Slightly confused why Travis hasn't picked this up already ...

hesselink:
I have no idea. Maybe because the master build is failing?

tomjaguarpaw:
The master branch is actually succeeding.  The "failing" status in the README comes from a different branch, which is rather misleading.

I'll have to look into this later.

tomjaguarpaw:
Closed and reopened to trigger Travis. Looks like a transient bug.

Issue #143, opened 2016-03-21
Support for JSON functions and operators in Postgres


drwebb:
I'm just learning Opaleye, and liking it so far. In a project I'm working on I'm considering using Postgres as a JSON document store. As an example, assume I have a table with two columns: `id :: integer, data :: jsonb`. Opaleye support for inserting `Values` into the `jsonb` column is working as expected. However, in recent versions of Postgres (>9.4) you are able to use operators on the JSON types for example: `SELECT* FROM posts WHERE data -> 'postPriority' > '50'`. `postPriority` is a attribute of the json objects stored in the data column.

This sounds like a nice enhancement to the Opaleye's Query language. It would be nice to have some set of the operators defined in
http://www.postgresql.org/docs/9.5/static/functions-json.html,
once an GIN index is defined on the jsonb column as in
http://www.postgresql.org/docs/9.5/static/datatype-json.html.

Any thoughts?

tomjaguarpaw:
They'd be fairly easy to add, with the caveat that they all (or at least most) produce nullable results making things a little bit fiddly.

tomjaguarpaw:
(Fiddly to use that is, not to add to Opaleye)

drwebb:
That seems reasonable, a `jsonb` is going right into schema-less database territory. We're still evaluating our needs for future flexibility, i.e. do we want a `jsonb` column or to just go with a relational model. If we decide on the JSON route I'll be happy to take a stab at this.

tomjaguarpaw:
If you take a stab feel free to contact me if I can be of any assistance.

tomjaguarpaw:
Hi @drwebb, have you seen the changes in https://github.com/tomjaguarpaw/haskell-opaleye/issues/193?  They may be useful for you.

tomjaguarpaw:
Closing as stale and probably fixed.

Issue #156, opened 2016-04-20
Support for tstzrange


lippling:
Is there any support for the `tstzrange` type?

(We use temporal tables - http://clarkdave.net/2015/02/historical-records-with-postgresql-and-temporal-tables-and-sql-2011/)

tomjaguarpaw:
I don't think so but it doesn't sound hard to add.  You can add a type here

https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/PGTypes.hs

and perhaps functions for working with them like `timestamptzAtTimeZone` here

https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Operators.hs

If you want to run queries that return them, an `instance QueryRunnerColumnDefault` needs to be added here

https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/RunQuery.hs

(You can only run queries if postgresql-simple already supports them.  I'm not sure if it does.)

lippling:
Haven't checked it, but there seems to be some support (https://hackage.haskell.org/package/postgresql-simple-0.5.1.2/docs/Database-PostgreSQL-Simple-Range.html).

What would be the recommended approach (roughly) to add operators like http://www.postgresql.org/docs/9.5/static/functions-range.html#RANGE-OPERATORS-TABLE for the range type(s)?

tomjaguarpaw:
You should probably implement them like `.<`

https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Operators.hs#L76

Firstly you'll add a constructor in `Opaleye.Internal.HaskellDB.PrimQuery` and then chase through all the things that that implies (SQL generation etc.).  The type checker should point you to the right places.

lippling:
Thanks!
I'll give it a try in the next couple of weeks.

xldenis:
This is fixed by #221 and this issue should be closed.

tomjaguarpaw:
Thanks!

PR #193, opened 2016-08-24
Add json operators

roelvandijk:
Partially implements #143.

Adds the operators specified in https://www.postgresql.org/docs/9.4/static/functions-json.html.

Note that `.@>`, `.<@`, `.?`, `.?|` and `.?&` have a stricter type than specified by PosgreSQL. In PostgreSQL their operands may be `null`. I think the not null variants are more useful in practice.

In order to overload `.->` and `.->>` I created two type classes. A user could easily make a type like `PGInt4` an instance of `IsPGJson` allowing queries to be written that crash at runtime. To prevent that I used the private super class trick.


tomjaguarpaw:
This looks great!  Could you add some tests please?

roelvandijk:
Sure! I added some tests.

roelvandijk:
Could the Travis test failure be the result of an older version of Postgresql which does not yet have the `jsonb` type?

tomjaguarpaw:
It uses Postgres 9.3, so you are correct, it does not have `jsonb`.  Let's just skip `jsonb` checks for now, as they are exactly the same as the `json` ones.

roelvandijk:
Ok, I'll comment out the jsonb tests and force push the result.

tomjaguarpaw:
It would be good if you could add them to another list of tests which is compiled but not run.  Then we can easily enable them in the future.

roelvandijk:
Good idea, I'll do that

roelvandijk:
Oops. Forgot the table creation functions

roelvandijk:
Yay, finally builds!

I realized I can also make the jsonb testing code dependent on **not** running on travis.

Something like:
```haskell
jsonbResults <- case travis' of 
  True -> []
  False -> do
    mapM_ (PGS.execute_ conn . dropAndCreateTableJsonb) jsonbTables
    insert (table9, table9columndata)
    mapM ($ conn) jsonbTests
```

Would that be useful?

roelvandijk:
Ok, last idea for today. What are you thoughts on adding the json operators to the `PrimQuery` language?

I would be happy to add some constructors to `BinOp` and some SQL translations to `showBinOp`.

Something like:
```haskell
data BinOp = ...
           | OpJsonGet              -- "->"
           | OpJsonGetText          -- "->>"
           | OpJsonGetPath          -- "#>"
           | OpJsonGetPathText      -- "#>>"
           | OpJsonbLeftWithinRight -- "@>"
           | OpJsonbRightWithinLeft -- "<@"
           | OpJsonbContains        -- "?"
           | OpJsonbContainsAny     -- "?|"
           | OpJsonbContainsAll     -- "?&"
```

tomjaguarpaw:
I think it's a good idea.  The only downside is that the names don't correspond to the symbols at all.  I'll leave it up to you.  If you want to do it, please do.  If not, no problem.

roelvandijk:
The naming part is hard. An alternative is to use symbols as names.

Simply use `:.->` or `:->` as the constructor corresponding to `.->`. Likewise for the other operators.

A downside to using symbols is that it is not symmetrical with the other operator constructors. Unless you also change them to symbols.

tomjaguarpaw:
Using type operators is a clever idea!  Please submit a separate PR and I'll have a look at it.

roelvandijk:
If #194 is appealing I can do the same for the json operators.

roelvandijk:
I fixed a typo. Now the test suite is happy again.

If necessary I can rebase all 3 commits into a single commit.

I noticed that there is some minor overlap possible with #156. Namely the `<@` and `@>` operators are also used by [ranges](https://www.postgresql.org/docs/9.5/static/functions-range.html#RANGE-OPERATORS-TABLE). This can be solved in time using overloading. Same as for `+`, `-`, `*` and friends (who are also defined for ranges).

tomjaguarpaw:
This looks fantastic.  I have one last group of requests:

* Can you rename `IsPGJson` to `PGJson`.  We already have `PGNum`, `PGFractional` etc..
* Can you remove `InternalIsPGJson` and just use `PGJson`?  I agree with your rationale but we don't use the hidden superclass trick for any other of our typeclasses so let's be consistent and not use it here either.

roelvandijk:
I didn't think of the relation with the existing classes. Good call.

I'll add a small warning to the `PGJson` and `JsonIndex` classes to discourage users from making instances.

Also `PGJsonIndex` instead of `JsonIndex` perhaps? It is the class of types which can be used as json keys or indexes in PostgreSQL.

tomjaguarpaw:
> I'll add a small warning

Good idea.

> Also PGJsonIndex instead of JsonIndex perhaps?

And another good idea.

roelvandijk:
Problem: `PGJson` is already the name of a type. This conflicts with the proposed class name.

```haskell
data PGJson
data PGJsonb
```
Need to think of another name for our bikeshed :-)

```haskell
class PGJsonLike a
```
?

tomjaguarpaw:
That's fine, or you could use `PGIsJson` :)

tomjaguarpaw:
It's up to you.  I'll be happy with either of those.

roelvandijk:
Done. I went with `PGIsJson` :-)

Would you like the commits rebased into a single commit?

tomjaguarpaw:
Yes, please rebase and I'll merge after it goes through Travis.

tomjaguarpaw:
Rebased this here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/201

tomjaguarpaw:
Merged this here

https://github.com/tomjaguarpaw/haskell-opaleye/pull/201

PR #194, opened 2016-08-24
Mimic operator names in BinOp constructors

roelvandijk:
`BinOp` constructors are renamed to syntactically match the corresponding PostgreSQL operators.

Some operators have names which are not valid Haskell operators (`AND`, `LIKE`, `MOD`). I did not changes names in those instances.

One outlier is the comparison operator `=`. I chose to represent it as `:==` in Haskell in order to distinguish it from the assignment operator `:=`. 

The advantage of these new names is that the DSL more closely matches the target SQL language.


tomjaguarpaw:
I like this!  One question: why `.==` instead of `.=`?

roelvandijk:
Because equality testing and assignment use the same operator in PostgreSQL.

tomjaguarpaw:
Fair enough.  I think this is a good idea.  The only thing that can go wrong is if two morally distinct operators have the same symbol (exactly as in the case of `=` and `==`) but I think we'll deal with that as we come to it.

roelvandijk:
I think operators having the same symbol will be very rare. It also leads to problems in PostgreSQL's grammar.

tomjaguarpaw:
Cool. I'm happy to accept this.  I'm rerunning one of the tests since I think it had a spurious failure.

tomjaguarpaw:
This is very cool, thanks!

roelvandijk:
No problem. I'm enjoying using opaleye.

Issue #197, opened 2016-09-03
Request for docs: How to handle ENUM?


saurabhnanda:
Couldn't find how to handle ENUMs in the tutorial docs. How do I make `tenant_status` in the following schema an ADT in Opaleye?

```
create type tenant_status as enum('active', 'inactive');
create table tenants(
       id serial primary key
       ,created_at timestamp without time zone not null default current_timestamp
       ,updated_at timestamp without time zone not null default current_timestamp
       ,name text not null
       ,status tenant_status not null default 'inactive'
       ,owner_id integer
       ,backoffice_domain text not null
       constraint ensure_not_null_owner_id check (tenent_status!='active' or owner_id is not null)
);
```

matsumonkie:
Say you have:
```
data Gender = Female | Male deriving (Eq, Show, Generic)

data User' a = User'
  {  usrGender :: a
  } deriving (Eq, Show, Generic)

type User = User' (Gender)
type RUserColumn = User' (Column PGText)
type WUserColumn = User' (Column PGText)

```

One way to have your usrGender attribute parsed from PG to your code is by having `Gender` override `FromField ` and `QueryRunnerColumnDefault`:

```
instance FromField (Gender) where
  fromField f mdata =
    return gender
    where
      gender = case mdata of
        Just "female" -> Female
        Just "male" -> Male
        _ -> Male

instance QueryRunnerColumnDefault PGText Gender where
  queryRunnerColumnDefault = fieldQueryRunnerColumn
```

So for your example, I guess you need to do the same with but with your enum type.
There are other [possible ways](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/RunQuery.hs#L121-L141) to deal with this case

saurabhnanda:
Thanks @matsumonkie this solved my problem.

matsumonkie:
You're welcome ;-)

saurabhnanda:
@matsumonkie @tomjaguarpaw do you know how to set-up the default Haskell => PG mapping for ENUMS? I managed to get the following:

```
instance ToField TenantStatus where
  toField ts = case ts of
    TenantActive -> toField "active"
    TenantInactive -> toField "inactive"
    TenantNew -> toField "new"
```

but I'm unable to figure out how to get Opalaye to use it:

```
activateTenant :: TenantId -> AppM (Tenant)
activateTenant tenantId@(TenantId tid) = do
[...]

PR #203, opened 2016-09-10
Documentation on db types to haskell types convertion

matsumonkie:
First thing first, thanks a lot for this awesome project ! 
I'm in the process the try to understand the source as I find it really interresting.

As a Haskell beginner, I sometime found a bit hard to use Opaleye. Issue #197 is something I have also ran into. I thought it could be nice to add some documentation on how to handle convertions.

What do you think ?

### DIFF
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -3,6 +3,7 @@
 > {-# LANGUAGE FlexibleInstances #-}
 > {-# LANGUAGE MultiParamTypeClasses #-}
 > {-# LANGUAGE TemplateHaskell #-}
+> {-# LANGUAGE OverloadedStrings #-}
 >
 > module TutorialBasic where
 >
@@ -17,6 +18,8 @@
 >                          showSqlForPostgres, Unpackspec,
 >                          PGInt4, PGInt8, PGText, PGDate, PGFloat8, PGBool)
 >
+> import           Opaleye.RunQuery
+>
 > import           Data.Profunctor.Product (p2, p3)
 > import           Data.Profunctor.Product.Default (Default)
 > import           Data.Profunctor.Product.TH (makeAdaptorAndInstance)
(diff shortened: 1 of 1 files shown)


saurabhnanda:
Why does one need to derive `FromField` AND `QueryRunnerColumnDefault`? Looking at the API surface, it seems that the latter has all the information required to teach Opaleye how to convert a `PGText` to `Gender`:

```instance QueryRunnerColumnDefault PGText Gender where```

Also, what about the other way round? How does Opaleye know how to convert a `Gender` to `PGText` (or whatever else raw SQL is going to accept)?

saurabhnanda:
Also, I'm not clear how `PGText` came into the picture? In case of an `ENUM` datatype, wouldn't Postgres send it as a completely new type, altogether?

saurabhnanda:
Also, a link to https://hackage.haskell.org/package/postgresql-simple-0.5.2.1/docs/Database-PostgreSQL-Simple-FromField.html in the documentation might be a good idea. It contains a more detailed example of how to handle parsing failures as well.

matsumonkie:
> Also, I'm not clear how PGText came into the picture? In case of an ENUM datatype, wouldn't Postgres send it as a completely new type, altogether?

I think Postgres handle an enum type internally as an Int but present it as a string. Your enum type should probably as a PGString, not sure though.

> Also, a link to https://hackage.haskell.org/package/postgresql-simple-0.5.2.1/docs/Database PostgreSQL-Simple-FromField.html in the documentation might be a good idea. It contains a more detailed example of how to handle parsing failures as well.

Yeah you're right, I have changed the PR. Thanks

> Also, what about the other way round? How does Opaleye know how to convert a Gender to PGText (or whatever else raw SQL is going to accept)?

I don't know yet, still learning Opaleye :-)

tomjaguarpaw:
Thanks very much for this.  I don't really have time at the moment to handle all the inbound Opaleye requests I'm getting but I will get to this sooner or later!

PR #208, opened 2016-09-17
defaultSqlExpr: Fix handling of `BinExpr OpIn _ (ListExpr _)`

k0001:
There's an issue in how `BinExpr OpIn _ (ListExpr _)` expressions are
currently handled in `defaultSqlExpr`. For example, consider this
`PrimExpr`:

```haskell
pe1 :: PrimExpr
pe1 = BinExpr OpIn
              (ConstExpr (BoolLit True))
              (ListExpr [ConstExpr (BoolLit True),
                         ConstExpr (BoolLit False)])
```

Before this change, we get redundant parentheses, which makes PostgreSQL
treat `((TRUE,FALSE))` as a list of records, instead of a list of
booleans as expected:

```
> ppSqlExpr (defaultSqlExpr defaultSqlGenerator pe1)
TRUE IN ((TRUE,FALSE))
```

After this change, we get the list of booleans as expected:

```haskell
> ppSqlExpr (defaultSqlExpr defaultSqlGenerator pe1)
TRUE IN (TRUE,FALSE)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## HEAD
+
+* Fixed handling of `BinExpr OpIn _ (ListExpr _)` in `defaultSqlExpr`.
+
 ## 0.5.1.0
 
 * Added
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Cool, thanks for spotting that.  I'm not sure how this slipped past the tests.  I'm away at the moment so this will have to wait until I get back.  Something funny is also going on with Cabal and it fails Travis.

k0001:
@tomjaguarpaw Thanks for the quick reply. Travis is failing for GHC HEAD, it builds fine for the other GHC versions.

PR #209, opened 2016-09-17
Opaleye.Operators.in_ now actually uses the SQL `IN` operator.

k0001:
This change (a0828ba3caff6a14835fd06fba89863c9954e53e) is backwards compatible, and builds on top of #208.

Example: Using `in_` with a non-empty foldable uses `OpIn`.

```haskell
> in_ [pgBool True, pgBool False] (pgBool True)
Column (BinExpr OpIn
                (ConstExpr (BoolLit True))
                (ListExpr [ConstExpr (BoolLit True),
                           ConstExpr (BoolLit False)]))
```

Example: Using `in_` with an empty foldable is just `False`.

```haskell
> in_ [] (pgBool True)
Column (ConstExpr (BoolLit False))
```

A test named `testIn` was added to ensure `in_` behaves as expected,
both when given empty and non-empty `Foldable`s. The SQL output of the
query in the test is:

```sql
SELECT "column10_1" as "result1_2",
       "column21_1" as "result2_2"
FROM (SELECT *
      FROM (SELECT "column1" as "column10_1",
                   "column2" as "column21_1"
            FROM "table1" as "T1") as "T1"
      WHERE (NOT (FALSE)) AND ("column21_1" IN (100,200))) as "T1"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## HEAD
+
+* Fixed handling of `BinExpr OpIn _ (ListExpr _)` in `defaultSqlExpr`.
+
 ## 0.5.1.0
 
 * Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## HEAD
 
 * Fixed handling of `BinExpr OpIn _ (ListExpr _)` in `defaultSqlExpr`.
+* `in_` now actually uses the SQL `IN` operator.
 
 ## 0.5.1.0
 
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -346,6 +346,16 @@ testRestrict = testG query
           O.restrict -< fst t .== 1
           Arr.returnA -< t
 
+testIn :: Test
+testIn = testG query expected
+  where query = proc () -> do
+          t <- table1Q -< ()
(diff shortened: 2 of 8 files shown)


tomjaguarpaw:
Seems like a fine idea, but is there something wrong with the existing implementation?

tomjaguarpaw:
While you are there, could you make the `ListExpr` constructor take a non-empty list?

tomjaguarpaw:
In fact `ListSqlExpr` should be changed and then all the other needed changes will propagate from there.

k0001:
>  Seems like a fine idea, but is there something wrong with the existing implementation?

I don't think so, other than the fact that this approach generates SQL that's easier to read if required.

k0001:
In https://github.com/tomjaguarpaw/haskell-opaleye/pull/209/commits/7d9d87ea4842a244eb85e73f2fa6a987b9599bca I made the `Data.List.NonEmpty` change you suggested.

tomjaguarpaw:
Very nice.  If you can fix the Cabal build error in HEAD then I can merge it from here, otherwise it will have to wait until I get back.

k0001:
@tomjaguarpaw all fixed!

tomjaguarpaw:
Super

k0001:
@tomjaguarpaw thanks!

PR #210, opened 2016-09-18
Tests: allow specifying the env variable OPALEYE_TEST_PG_CONN

k0001:
Currently, running the test suite is a rather painful experience because one has to modify the `Test/Test.hs` every time in order to specify the PostgreSQL connection details (and then undo those changes before pushing anything upstream, in case one is making changes to the tests themselves). 

This is a backwards compatible change that allows specifying the `OPALEYE_TEST_PG_CONN` environment variable when running the tests, where one can specify the correct PostgreSQL connection string to use.

Example usage:

``` bash
OPALEYE_TEST_PG_CONN=postgresql://opa:opa@localhost/opa-test runghc ./Setup.hs test
```

If `OPALEYE_TEST_PG_CONN` is not specified, then the behavior is the same as it was before: Use the Travis CI settings if running in Travis CI, otherwise use the fallback defaults that are hardcoded in the `Test/Test.hs` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Fixed handling of `BinExpr OpIn _ (ListExpr _)` in `defaultSqlExpr`.
 * `in_` now actually uses the SQL `IN` operator.
+* Tests: Allow specifing the PostgreSQL connection string via the
+  `OPALEYE_TEST_PG_CONN` environment variable.
 
 ## 0.5.1.0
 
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
Thanks, it's a good idea to make the test suite configurable, but I am very much not a fan of environment variables being, as they are, implicit global state.  Is there no more Haskelly way of doing this?

k0001:
@tomjaguarpaw I think this is the right approach, seeing as there is one standard way of running tests `(i.e., `runghc ./Setup.hs test`, also known as `cabal test` or `stack test`, which doesn't take any command-line arguments), and we should aim to play nice with that. Travis CI encourages this approach as well.

Notice that nothing goes wrong if environment variables are not set.

tomjaguarpaw:
I eventually got round to merging this, which does much the same thing https://github.com/tomjaguarpaw/haskell-opaleye/pull/287

PR #221, opened 2016-10-28
Add Postgres Ranges

xldenis:
This PR attempts to implement postgres range types, this should solve https://github.com/tomjaguarpaw/haskell-opaleye/issues/156. There is still a lot of work to do but I figured I would open it up to gather opinions and get CI working. 

**notes:** tests seem very incomplete. I also can't seem to get the warning for non-exhaustive-patterns which makes it hard to know which methods need additional cases. 

Ranges have four possible conditions, documented [here](https://hackage.haskell.org/package/postgresql-simple-0.5.1.2/docs/Database-PostgreSQL-Simple-Range.html). I wasn't sure what the best approach was so I used the data type defined in `postgresql-simple`. 

I've added the operators defined [here](https://www.postgresql.org/docs/9.6/static/functions-range.html).


tomjaguarpaw:
This is great, thanks!  What problems are you having with non-exhaustive-patterns?

xldenis:
> What problems are you having with non-exhaustive-patterns?

Example: When I added the new operators I got zero warning about the missing cases in `showBinOp`

tomjaguarpaw:
That's strange.  [`-Wall` is on](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/opaleye.cabal#L93).  Shouldn't that warn on incomplete patterns?  (In the back of my mind I have a feeling I've seen this issue before.)

tomjaguarpaw:
Would you say that it's just tests missing for this now, or are you planning to add something else?

xldenis:
> That's strange. -Wall is on. Shouldn't that warn on incomplete patterns? (In the back of my mind I have a feeling I've seen this issue before.)

It should which is why I'm so perplexed.. I thought it was because it wasn't recompiling the relevant modules but that doesn't seem to be it either..

To be entirely fair, the output from `stack build` is quite different on opaleye than on other projects, it doesnt seem to show the results of each module compilation (maybe it's hiding the logs?).

xldenis:
> Would you say that it's just tests missing for this now, or are you planning to add something else?

Assuming there are no partial functions left, it should be just tests!

xldenis:
> According to the tests there is a missing pattern!

I'd still like to figure out why I can't seem to get that output during builds 😕 . But at least the test output helps point that out!

tomjaguarpaw:
Yeah, it's weird.

bergmark:
Did you try stack clean && stack build --pedantic?

xldenis:
Aha! I've found it out. `stack build` builds opaleye as a dependency of the overall project (and then builds opaleye-sqlite. Running `stack build opaleye` fixes the issue!

tomjaguarpaw:
There's a test failure but I don't have much clue where it arises from:

https://travis-ci.org/tomjaguarpaw/haskell-opaleye/jobs/171904898

xldenis:
I think I know what causes it but I need to run the tests locally to confirm.

tomjaguarpaw:
Any idea what's going on here?

lippling:
`src/Opaleye/Internal/HaskellDB/Sql/Print.hs:150` has a parse error.

tomjaguarpaw:
@lippling Well spotted!

@xldenis Can you fix `src/Opaleye/Internal/HaskellDB/Sql/Print.hs:150`?  Then hopefully it will be good to merge.

lippling:
I'm on my phone right now: does this PR also add support for `@>`, `<@` with `tstzrange`?

tomjaguarpaw:
Not as far as I can see, but I'm sure we can do something about that.

lippling:
(That refers to #156)

xldenis:
Not explicitly. I haven't tested to see if Opaleye generates valid sql for those.

xldenis:
I've fixed up the tests. Postgres doesn't like interacting with literal ranges, maybe the SQL should always cast them to their type (since we know it from Haskell)?

tomjaguarpaw:
Yes, as long as we do always know the type.  How do you distinguish between `int4range`s and other sorts of ranges?

xldenis:
> How do you distinguish between int4ranges and other sorts of ranges?

There is a set of default defined range types. I've added a `IsRangeType` class to distinguish the set of types that are valid range types. It provides a function which could return the name of the range type. Custom types would have to implement the `IsRangeType` class to leverage the range operators.

tomjaguarpaw:
I'm a bit confused then.  Why are you not actually using `showRangeType` when constructing the ranges with `pgRange`?

xldenis:
> I'm a bit confused then. Why are you not actually using showRangeType when constructing the ranges with pgRange?

I haven't added the cast for range literals yet. I'll add that today and this should be good to ship!

tomjaguarpaw:
Fantastic work!

tomjaguarpaw:
QuickCheck has been disabled!

src/Opaleye/PGTypes.hs
```diff
@@ -178,6 +187,29 @@ instance IsSqlType PGJson where
   showPGType _ = "json"
 instance IsSqlType PGJsonb where
   showPGType _ = "jsonb"
+instance IsRangeType a => IsSqlType (PGRange a) where
+  showPGType _ = showRangeType ([] :: [a])
```
tomjaguarpaw: Can this not be

````haskell
showPGType = showRangeType
````

?

File: Test/Test.hs [hunk 2]
```diff
@@ -912,7 +951,7 @@ main = do
   -- insert (table9, table9columndata)
 
   -- Need to run quickcheck after table data has been inserted
-  QuickCheck.run conn
+  -- QuickCheck.run conn
```
tomjaguarpaw: Please reenable!

lippling:
I just played a bit with the PR.

To be specific:

I tried to add `r @> e` for my use case ("is element e in range r?"), but this operator is already defined in Opaleye (for JSON). So PSQL overloads some operators. Question is: how can we deal with that?

For now, I defined
```haskell
infix 4 ..@>
(..@>) :: Column (T.PGRange a) -> Column a -> Column T.PGBool
(..@>) = C.binOp (HPQ.:@>)
```

which seems to work without further changes.

tomjaguarpaw:
For this case I don't like the idea of overloading because the two use cases
are completely different.

Let's go with your suggestion of a new operator name.

lippling:
Number one problem in IT: Naming...

src/Opaleye/PGTypes.hs
xldenis: yep! that would be more correct.

xldenis:
The overloading of operators is why I had to add a cast for all range literals since psql can't infer the type of `'[2.0,3.0]'` on it's own!

xldenis:
I actually don't agree with the creation of an operator for the range.. In my opinion they serve the same role in both cases. `a @> b` is just saying does `a` contain `b`. However, they shouldn't be defined over all columns. I think the correct thing would be to place a constraint on the types can be used with `.@>` maybe a class `Iterable` ? (name is subject to change ofc). This is an operator that should apply to _all_ 'enumerable' psql types.

tomjaguarpaw:
Oh, so you're saying it's genuinely the same operator?  Hmm ... well I don't terribly mind.  Let's see if we can work out a good proposal between us.

xldenis:
Yep the alias in all cases for it is `contains`, its the same operation over different structures.

tomjaguarpaw:
I squashed this and opened a new PR here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/240

tomjaguarpaw:
Merged here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/240

tomjaguarpaw:
Thanks @xldenis!

PR #224, opened 2016-11-01
Extract test configuration to yml

xldenis:
In order to make testing in different environments simpler, I've extracted the configuration for tests to external YML files. This lets me more easily change my configuration to run tests locally. 

I'll probably completely remove `config.yml` and only provide `config.default.yml` as an example configuration while adding `config.yml` to `.gitignore`. 

I'm not quite sure why travis can't seem to find `config.travis.yml`, running tests locally with `TRAVIS=yes` doesn't seem to cause any issues.

Ah it seems like the CI script builds the code into a different directory but doesnt include the configuration file. I guess the proper solution would be something similar to `dotenv`.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE Arrows #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
-
+{-# LANGUAGE DeriveGeneric #-}
 module Main where
 
 import qualified QuickCheck
@@ -20,6 +20,7 @@ import           Data.Monoid ((<>))
 import qualified Data.String as String
 import qualified Data.Time   as Time
 import qualified Data.Aeson as Json
+import qualified Data.Yaml  as Yaml
 import qualified Data.Text as T
 
 import qualified System.Exit as Exit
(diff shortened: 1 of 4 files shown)


tomjaguarpaw:
I'm definitely in favour of making the testing code more configurable.  @k0001 made another attempt here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/210/files

However, I'm going to be very conservative about this because it's a very important piece of the Opaleye infrastructure and I don't want to get bogged down in changes to it.  Currently it's very crude but it's simple and effective.

If anyone wants to share examples of best practices regarding testing configuration from other projects I would welcome it.

xldenis:
>However, I'm going to be very conservative about this because it's a very important piece of the Opaleye infrastructure and I don't want to get bogged down in changes to it. Currently it's very crude but it's simple and effective.

It is definitely crude :), however, I think its _very_ important to improve it. I think it is very discouraging for other developers to pitch in and run the test suite. 

 I've made basic changes so the development environment no longer requires changing `Test.hs`. However, running tests from CI is unaffected.

Ideally, something akin to the `dotenv` pattern could be used, read values from config and/or environment simultaneously. 

Finally, I'd like to take a stab at fixing up the test suite. Right now there is no information that helps identify _which_ test actually failed. There is a lot of work that can be done on that front.

tomjaguarpaw:
> I think it is very discouraging for other developers to pitch in and run the test suite. 

Point taken.  I don't see why changing `Test.hs` is so discouraging though.  Is it because it interacts badly with version control?

> Finally, I'd like to take a stab at fixing up the test suite. Right now there is no information that helps identify which test actually failed. There is a lot of work that can be done on that front. 

That would be great!  Converting to Tasty/HSpec would be really useful.

xldenis:
>  I don't see why changing Test.hs is so discouraging though. Is it because it interacts badly with version control?

Exactly! The second bit is that it isn't possible to know what the failing test case is (major minus imo). 

> Converting to Tasty/HSpec would be really useful.

I'd like to convert to use `hspec-discover`.

tomjaguarpaw:
On Sun, Nov 06, 2016 at 04:20:43PM -0800, Xavier Denis wrote:
> > Converting to Tasty/HSpec would be really useful.
> 
> I'd like to convert to use `hspec-discover`. 

Via Tasty?  Or ignoring Tasty entirely?

xldenis:
Ignoring it entirely, I haven't had much of a use for it when using `hspec` combined with `hspec-discover`.

tomjaguarpaw:
On Sun, Nov 06, 2016 at 06:43:17PM -0800, Xavier Denis wrote:
> Ignoring it entirely, I haven't had much of a use for it when using `hspec` combined with `hspec-discover`. 

OK, that's fine!  Looking forward to seeing what you come up with.

xldenis:
for ref: https://github.com/tomjaguarpaw/haskell-opaleye/pull/236

By extracting the connection to a yml file, it can be extended to include other test configuration such as whether to run `QuickCheck` which can be nice to turn off during development.

tomjaguarpaw:
Closing this now that we have https://github.com/tomjaguarpaw/haskell-opaleye/pull/287 instead

PR #236, opened 2016-11-12
Hspec tests

xldenis:
I've refactored the test suite to use HSpec.

This is still very much WIP, however I figured I'd start gathering opinions on the changes now. I am pleasantly surprised both by how easy it was to refactor and how it has enabled some miscellaneous cleanups!

TODO:

- [x] Re-enable Quickcheck
- [x] fix configuration

A followup PR will split `Test.hs` into submodules containing relevant suites (ie ranges, json, arrays, etc) and then we can leverage `hspec-discover` to build the test listing automatically.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -33,6 +33,7 @@ import           Control.Arrow ((&&&), (***), (<<<), (>>>))
 
 import           GHC.Int (Int64)
 
+import Test.Hspec
 -- { Set your test database info here.  Then invoke the 'main'
 --   function to run the tests, or just use 'cabal test'.  The test
 --   database must already exist and the test user must have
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Thanks very much for doing this.  It might take me a little while before I can get to your PRs but I really appreciate you doing them.

Test/Test.hs
```diff
@@ -32,16 +32,17 @@ import           Control.Arrow ((&&&), (***), (<<<), (>>>))
 
 import           GHC.Int (Int64)
 
+import Test.Hspec
 -- { Set your test database info here.  Then invoke the 'main'
---   function to run the tests, or just use 'cabal test'.  The test
+--   function to run the tests, or just use 'cabal Test.  The test
```
xldenis: lol woops `s/test'/test/i`

Test/Test.hs
```diff
@@ -32,16 +32,17 @@ import           Control.Arrow ((&&&), (***), (<<<), (>>>))
...
 --   database must already exist and the test user must have
 --   permissions to modify it.
 
 connectInfo :: PGS.ConnectInfo
 connectInfo =  PGS.ConnectInfo { PGS.connectHost = "localhost"
-                               , PGS.connectPort = 25433
```
xldenis: obviously this needs to be reverted before merging.

xldenis:
> It might take me a little while before I can get to your PRs but I really appreciate you doing them.

No rush! This one was fairly mechanical once I figured out the type transformation.

xldenis:
I have some more changes that actually split up `Test.hs` into smaller submodules (and extracts a common `TestHelper` module for infrastructure / helper methods). Should I do those as part of this PR or should that be a follow up?

tomjaguarpaw:
> Should I do those as part of this PR or should that be a follow up?

Do those as a second PR so I can see more easily what's going on.

> Also, what are your overall feelings on this PR?

I think it's a great idea!  Sorry I haven't done anything with it yet.  I've been extremely busy.

xldenis:
I'm just adding a bump comment so we can merge this when you get back.

Test/Test.hs
tomjaguarpaw: Does this still need reverting?

Test/Test.hs
```diff
@@ -966,14 +936,82 @@ main = do
...
-  results <- mapM ($ conn) allTests
-
-  print results
-
-  let passed = and results
-
-  putStrLn (if passed then "All passed" else "Failure")
-  Exit.exitWith (if passed then Exit.ExitSuccess
-                           else Exit.ExitFailure 1)
+  -- QuickCheck.run conn
```
tomjaguarpaw: Need to reenable QuickCheck, presumably?

Test/Test.hs
tomjaguarpaw: Does this still need fixing?

Test/Test.hs
```diff
@@ -319,61 +320,64 @@ dropAndCreateDB conn = do
...
          -> PGS.Connection
-         -> IO b
-testG q p conn = do
+         -> Expectation
+
+testH q p conn = do
+  result <- O.runQuery conn q
+  p result
+
+queryShouldReturnSorted :: (D.Default O.QueryRunner wires haskells, Show haskells, Ord haskells) =>
```
tomjaguarpaw: Can't you do this in terms of `testH`?

Test/Test.hs
```diff
@@ -319,61 +320,64 @@ dropAndCreateDB conn = do
...
         -- executeJsonb = PGS.execute_ conn . dropAndCreateTableJsonb
 
-type Test = PGS.Connection -> IO Bool
+type Test = SpecWith PGS.Connection
 
-testG :: D.Default O.QueryRunner wires haskells =>
+testH :: D.Default O.QueryRunner wires haskells =>
          Query wires
-         -> ([haskells] -> b)
+         -> ([haskells] -> Expectation)
```
tomjaguarpaw: Why not `([haskells] -> expectation)`?

Test/Test.hs
xldenis: yes. I disable it during development because it takes significantly longer to run than the rest of the test suite.

Test/Test.hs
xldenis: Yes. Though ideally we can add non-source level configuration soon 😄

xldenis:
I've reverted the configuration modifications and re-enabled `QuickCheck`. I'll squash before a merge but leave it as is for any further fixups.

Test/Test.hs
```diff
@@ -33,8 +33,9 @@ import           Control.Arrow ((&&&), (***), (<<<), (>>>))
 
 import           GHC.Int (Int64)
 
+import Test.Hspec
 -- { Set your test database info here.  Then invoke the 'main'
---   function to run the tests, or just use 'cabal test'.  The test
+--   function to run the tests, or just use 'cabal test.  The test
```
tomjaguarpaw: Missing a closing quote?

tomjaguarpaw:
Just a missing close quote to do I think, otherwise it looks all fine.

tomjaguarpaw:
Can you fix the quote and squash? Then I'll merge.

tomjaguarpaw:
This is great stuff!

Sorry I didn't merge it earlier but I didn't get a notification about the new commit.

PR #238, opened 2016-11-21
infixr 3 .&&

roelvandijk:
- Fixes regression caused by 4fb62fdfb995aa257f303b077047ca52f5f15009

### DIFF
--- a/src/Opaleye/Operators.hs
+++ b/src/Opaleye/Operators.hs
@@ -123,6 +123,8 @@ infixr 2 .||
 (.||) :: Column T.PGBool -> Column T.PGBool -> Column T.PGBool
 (.||) = C.binOp HPQ.OpOr
 
+infixr 3 .&&
+
 -- | Boolean and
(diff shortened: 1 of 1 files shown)


roelvandijk:
Note that the `.&&` operator in `Opaleye.Internal.Operators` has the correct fixity. Why not simply re-export that symbol?

Note that you can set the [`not-home` module attribute](https://www.haskell.org/haddock/doc/html/module-attributes.html) for internal modules if you want nice Haddock documentation that includes internal symbols.

tomjaguarpaw:
Ah, it was for Haddock that I did it.  I will revert that change and look into `not-home`.

tomjaguarpaw:
Merged your patch so you have a fixed repo to work with and then I'll do the `not-home` stuff.

tomjaguarpaw:
Thanks, I uploaded this fix as 0.5.2.2 to Hackage.

PR #247, opened 2016-12-01
Looks like we don't actually need the Unpackspec component of QueryRunnerColumnDefault

tomjaguarpaw:
Perhaps we should remove this?


tomjaguarpaw:
This was done by f1c2431f637cd37af9e13cc27bb5d3a4935a200d

PR #255, opened 2016-12-25
Remove opaleye sqlite package

xldenis:
This is a fairly radical change so I understand if you don't want to go through with it. 

I've gone ahead and removed the entire sqlite subtree since at multiple times comments indicating that it is no longer developed have been made. It also seems that the code suffers of heavy duplication of the psql implementation... Seemingly indicating that it could be better abstracted or would benefit from waiting for backpack? 

If anything we can have a discussion about the future of opaleye-sqlite (which I find to be a cool idea).


tomjaguarpaw:
Let's definitely *not* remove opaleye-sqlite!  It implicitly contains
important info about what needs to be cahnged in Opaleye to get ti to work
with sqlite.

xldenis:
Sorry for the delay in responding, I had some family issues over this holiday season. 

>  It implicitly contains
important info about what needs to be cahnged in Opaleye to get ti to work
with sqlite.

Since we have everything versioned in git that information wouldn't be lost, it would be a matter of pulling up the correct revision. It seems to me that `opaleye-sqlite` is for all intents and purposes dead-code which confusingly shares much of the same module names... Its not easy to tell where the codebases share things and where they diverge.

If sqlite is kept and development continues what I'd like to see is the ability to have the same code execute on either an sqlite _or_ a postgres database (ie be agnostic to the actual underlaying db)... Of course only the common subset of SQL could be used like that if you want to use JSON operators then you'd have to guarantee a Postgres connection at the type level.

ocharles:
I'm in favour of the idea in this pull request, but it would be better to split out `opaleye-sqlite` into a separate repository (preserving the history, this can be done with Git subtree tools). Right now I constantly end up opening SQLite files when I work on `opaleye`

tomjaguarpaw:
I'm not entirely sure why I made `opaleye-sqlite` a subdirectory.  Maybe because it was simply the most convenient thing. Anyway, @merijn has shown some interest in taking over opaleye-sqlite so he should be part of the discussion.

merijn:
I've only had limited time to actually get started with this, but hopefully should have some more time now. I definitely think `opaleye-sqlite` should move to a separate repository at some point, if it's going to useful. But before such a move it'd be best to first sort out how to keep opaleye-sqlite in sync with opaleye. Such as, is there any common code that could be shared between the sqlite and postgres versions to reduce duplication of maintenance, etc.

I'll have some more useful comments (hopefully?) after I've spend a bit more time seeing what needs fixing/updating in the sqlite version.

xldenis:
Should we close this PR and open an issue instead?

tomjaguarpaw:
Seems like a good idea to make it an issue.

@merijn: I suggest you check out a version of the repo from the commit that last touched opaleye-sqlite and then do a diff to see what differed between the two trees.  Then hopefully you can port the differences to the current opaleye.  Finally, we will be able to do another diff to see what code we can share.

xldenis:
I've created an issue.

PR #264, opened 2017-01-14
Derive schema from table definition 

ryskajakub:
Hi sorry for a bit of inactivity, I adapted this PR to changes in master and added the typefunction that will replace `TableColumn` with `Column` for better inference. If this is OK, we can this typefunction - probably the generic version you talk about - to template haskell that is generated `product-profunctors` along with adapter and `Default` instance.


xldenis:
Will this allow things like generating insert statements / validating table schema on boot?
I'd love to be able to make sure that my database matches the expected (implicit) schema!


On a related note I've always wondered if it would be possible to model things like migrations as a type-level list, such that the resulting list can be reflected and checked for equality with the table definition / schema. 

ie:

```
m1 = Migr [AddColumn "A"]
m2 = Migr [AddColumn "B"]

data T1 = T1 { a :: _, b :: _ } deriving ... -- table definition

migrate $ m1 `after` m2 -- somehow assert that result will have 2 columns a and b
```

ryskajakub:
Yes, It should allow it. During the table schema definition, user will specify all column info with `required` or `optional`. Then we can run many extractors, that will gather table info, be it for generating `CREATE` statements, querying and also checking that the schema in database is equivalent to the declared and in case it's not it can suggest some `ALTER`s .

xldenis:
>  in case it's not it can suggest some ALTERs .

I don't think this a good idea but it's not relevant right now anyways.

tomjaguarpaw:
Thanks for this @ryskajakub!  Can you rebase on master so that the tests pass?

ryskajakub:
I fixed all errors but the one which uses `Table`'s `Functor`. Why is this needed? It looks like this can be substituted by running the `queryTable` and then applying the operations. Apart, perhaps, the case when you want to "hide" some columns from querying from the table.

If that would be needed, the `Column` type would be needed to be inserted to the `TableProperties` during the `Table` declaration, not during the querying. Also, there would need to be added extra type to the `TableProperties`. 

Or it could be designed differently.

tomjaguarpaw:
Having looked at this again I think that the changes are too invasive.  I don't like the idea of adding `TableColumn`.  It's one of those things that complicates type inference.  Yes, we already have several such things but I don't want any more if we can avoid it!

Instead I suggest that we add another field to `TableProperties`:

````haskell
data TableProperties writerColumns viewColumns = TableProperties
   { tablePropertiesWriter :: Writer writerColumns viewColumns
   , tablePropertiesView   :: View viewColumns
   , tableColumns          :: [(ColumnName, ColumnType)]
 }

type ColumnName = String
type ColumnType = String
````

The we will have to restrict `required` and `optional` a little bit:

````haskell
required :: String -> TableProperties (Column a) (Column a)
required columnName = T.TableProperties
  (T.required columnName)
  (View (Column (HPQ.BaseTableAttrExpr columnName)))
  [(columnName, showPGType (Proxy :: Proxy a))]

optional :: String -> TableProperties (Maybe (Column a)) (Column a)
optional columnName = T.TableProperties
  (T.optional columnName)
  (View (Column (HPQ.BaseTableAttrExpr columnName)))
  [(columnName, showPGType (Proxy :: Proxy a))]
````

Then I'm pretty sure we can get all or most of the table creation info we need out of that.

PR #268, opened 2017-02-10
Stackage lts 7.19

WillSewell:
I wanted to make this compatible with `aeson-1.1.*`. In doing so I did a stackage lts upgrade because it was quite out of date. I also updated the relevant upper bounds.

I can remove the last commit if you would prefer (and you are happy with the tests passing).

### DIFF
--- a/opaleye-sqlite/src/Opaleye/SQLite/Internal/PackMap.hs
+++ b/opaleye-sqlite/src/Opaleye/SQLite/Internal/PackMap.hs
@@ -40,7 +40,7 @@ import qualified Data.Functor.Identity as I
 -- 'ProductProfunctor') in @s@ and @t@.  It is unclear at this point
 -- whether we want the same @Traversal@ laws to hold or not.  Our use
 -- cases may be much more general.
-data PackMap a b s t = PackMap (Applicative f =>
+data PackMap a b s t = PackMap (forall f. Applicative f =>
                                 (a -> f b) -> s -> f t)
 
 -- | Replaces the targeted occurences of @a@ in @s@ with @b@ (changing
--- a/opaleye-sqlite/opaleye-sqlite.cabal
+++ b/opaleye-sqlite/opaleye-sqlite.cabal
@@ -28,16 +28,16 @@ library
     , base16-bytestring   >= 0.1.1.6 && < 0.2
     , case-insensitive    >= 1.2     && < 1.3
     , bytestring          >= 0.10    && < 0.11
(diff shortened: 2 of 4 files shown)


opaleye-sqlite/src/Opaleye/SQLite/Internal/PackMap.hs
```diff
@@ -40,7 +40,7 @@ import qualified Data.Functor.Identity as I
 -- 'ProductProfunctor') in @s@ and @t@.  It is unclear at this point
 -- whether we want the same @Traversal@ laws to hold or not.  Our use
 -- cases may be much more general.
-data PackMap a b s t = PackMap (Applicative f =>
+data PackMap a b s t = PackMap (forall f. Applicative f =>
```
WillSewell: I'm not exactly sure which change in the compiler necessitated this change. However it appears to be backwards compatible, so I don't see a problem with it.

tomjaguarpaw: Yeah, that's something that had to change in Opaleye a while ago but I didn't fix the sqlite version at the same time.

tomjaguarpaw:
Thanks Will, this is great.

WillSewell:
No problem. Thanks for merging :) It would be good if this could be released soon, otherwise `opaleye` won't make it into Stackage nightly builds after LTS 8.0 is out: https://github.com/fpco/stackage/issues/2177#issuecomment-278841830.

tomjaguarpaw:
I uploaded this last night.  Is that sufficient?

https://hackage.haskell.org/package/opaleye-0.5.3.0

WillSewell:
Ah yes. I was looking for a new version in the GitHub repo, but it looks like you haven't pushed the commit where you bumped the version yet.

tomjaguarpaw:
Good point.  I'm continually stymied by `git push --tags` not actually pushing the commits that the tags refer to!

Anyway, the new version is now on github master.

opaleye-sqlite/src/Opaleye/SQLite/Internal/PackMap.hs
tomjaguarpaw: @WillSewell Are you actually using opaleye-sqlite?  I'm not sure how widely used it is.

WillSewell: No I'm not, but I thought I'd update the deps for it since I was updating the `stack.yaml` resolver version anyway.

PR #271, opened 2017-02-28
Add NOT EXISTS support to the internal query representation

### LINKED ISSUE #270: EXISTS << subquery >> support
I need access to the `EXISTS` operator, which takes a query as an argument and returns a boolean as to whether or not that query returns any rows. Falling back to a `LEFT JOIN` isn't enough, a `LEFT JOIN` results in a considerably worse plan (due to planner statistics):

```
EXPLAIN ANALYZE 
SELECT * 
FROM 
  "projectOrder" po, 
  "bomSnapshot" bs 
WHERE 
  po.status = 'in-progress' AND 
  "projectOrderID" = po.id AND 
  NOT EXISTS (SELECT true FROM "bomSnapshotHistory" WHERE "previousId" = bs.id);
                                                                             QUERY PLAN                                                                             
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Anti Join  (cost=3.88..268.01 rows=1240 width=643) (actual time=0.362..1.634 rows=1251 loops=1)
   Hash Cond: ((bs.id)::bigint = ("bomSnapshotHistory"."previousId")::bigint)
   ->  Nested Loop  (cost=0.06..259.81 rows=1250 width=643) (actual time=0.284..1.182 rows=1265 loops=1)
         ->  Seq Scan on "projectOrder" po  (cost=0.00..67.67 rows=47 width=450) (actual time=0.271..0.306 rows=47 loops=1)
               Filter: ((status)::text = 'in-progress'::text)
               Rows Removed by Filter: 1295
         ->  Index Scan using "bomSnapshot_projectOrderID_idx" on "bomSnapshot" bs  (cost=0.06..4.00 rows=30 width=193) (actual time=0.003..0.008 rows=27 loops=47)
               Index Cond: (("projectOrderID")::bigint = (po.id)::bigint)
   ->  Hash  (cost=2.84..2.84 rows=281 width=8) (actual time=0.070..0.070 rows=281 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 19kB
         ->  Seq Scan on "bomSnapshotHistory"  (cost=0.00..2.84 rows=281 width=8) (actual time=0.004..0.033 rows=281 loops=1)
 Planning time: 0.447 ms
 Execution time: 1.750 ms
```

vs.

```
EXPLAIN ANALYZE 
SELECT *
FROM 
  "projectOrder" po, 
  "bomSnapshot" bs 
LEFT JOIN "bomSnapshotHistory" bsh ON bsh."previousId" = bs.id 
WHERE 
  po.status = 'in-progress' AND 
  bsh.id IS NULL AND 
  "projectOrderID" = po.id;
                                                                             QUERY PLAN                                                                             
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Left Join  (cost=3.88..264.60 rows=1 width=667) (actual time=0.357..1.658 rows=1251 loops=1)
   Hash Cond: ((bs.id)::bigint = (bsh."previousId")::bigint)
   Filter: (bsh.id IS NULL)
   Rows Removed by Filter: 14
   ->  Nested Loop  (cost=0.06..259.81 rows=1250 width=643) (actual time=0.291..1.210 rows=1265 loops=1)
         ->  Seq Scan on "projectOrder" po  (cost=0.00..67.67 rows=47 width=450) (actual time=0.276..0.314 rows=47 loops=1)
               Filter: ((status)::text = 'in-progress'::text)
               Rows Removed by Filter: 1295
         ->  Index Scan using "bomSnapshot_projectOrderID_idx" on "bomSnapshot" bs  (cost=0.06..4.00 rows=30 width=193) (actual time=0.003..0.008 rows=27 loops=47)
               Index Cond: (("projectOrderID")::bigint = (po.id)::bigint)
   ->  Hash  (cost=2.84..2.84 rows=281 width=24) (actual time=0.059..0.059 rows=281 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 24kB
         ->  Seq Scan on "bomSnapshotHistory" bsh  (cost=0.00..2.84 rows=281 width=24) (actual time=0.003..0.017 rows=281 loops=1)
```

I'd add this myself, but it looks like it's going to be quite an intrusive change. My plan would be to add `Exists` to `PrimExpr`, with a `PrimQuery' ()` as an argument - but with the current module structuring that causes a cycle. I'm not sure how you want to resolve that.

ocharles:
Fixes #270

### DIFF
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -53,6 +53,7 @@ data PrimQuery' a = Unit
                               (Bindings HPQ.PrimExpr)
                               (PrimQuery' a)
                               (PrimQuery' a)
+                  | Antijoin  (PrimQuery' a) (PrimQuery' a)
                   | Values    [Symbol] (NEL.NonEmpty [HPQ.PrimExpr])
                   | Binary    BinOp
                               (Bindings (HPQ.PrimExpr, HPQ.PrimExpr))
@@ -79,6 +80,7 @@ data PrimQueryFold' a p = PrimQueryFold
               -> p
               -> p
               -> p
+  , antijoin  :: p -> p -> p
   , values    :: [Symbol] -> (NEL.NonEmpty [HPQ.PrimExpr]) -> p
   , binary    :: BinOp -> (Bindings (HPQ.PrimExpr, HPQ.PrimExpr)) -> (p, p) -> p
   , label     :: String -> p -> p
@@ -101,6 +103,7 @@ primQueryFoldDefault = PrimQueryFold
   , binary    = Binary
   , label     = Label
   , relExpr   = RelExpr
+  , antijoin  = Antijoin
   }
 
 foldPrimQuery :: PrimQueryFold' a p -> PrimQuery' a -> p
@@ -118,6 +121,7 @@ foldPrimQuery f = fix fold
           Binary binop pes (q1, q2) -> binary    f binop pes (self q1, self q2)
           Label l pq                -> label     f l (self pq)
           RelExpr pe syms           -> relExpr   f pe syms
+          Antijoin q1 q2            -> antijoin  f (self q1) (self q2)
         fix g = let x = g x in x
 
 times :: PrimQuery -> PrimQuery -> PrimQuery
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -53,7 +53,7 @@ data PrimQuery' a = Unit
                               (Bindings HPQ.PrimExpr)
                               (PrimQuery' a)
                               (PrimQuery' a)
-                  | Antijoin  (PrimQuery' a) (PrimQuery' a)
+                  | Exists    Bool (PrimQuery' a) (PrimQuery' a)
                   | Values    [Symbol] (NEL.NonEmpty [HPQ.PrimExpr])
                   | Binary    BinOp
                               (Bindings (HPQ.PrimExpr, HPQ.PrimExpr))
@@ -80,7 +80,7 @@ data PrimQueryFold' a p = PrimQueryFold
               -> p
               -> p
               -> p
-  , antijoin  :: p -> p -> p
+  , existsf   :: Bool -> p -> p -> p
   , values    :: [Symbol] -> (NEL.NonEmpty [HPQ.PrimExpr]) -> p
   , binary    :: BinOp -> (Bindings (HPQ.PrimExpr, HPQ.PrimExpr)) -> (p, p) -> p
   , label     :: String -> p -> p
@@ -103,7 +103,7 @@ primQueryFoldDefault = PrimQueryFold
   , binary    = Binary
   , label     = Label
   , relExpr   = RelExpr
-  , antijoin  = Antijoin
+  , existsf   = Exists
   }
 
 foldPrimQuery :: PrimQueryFold' a p -> PrimQuery' a -> p
@@ -121,7 +121,7 @@ foldPrimQuery f = fix fold
           Binary binop pes (q1, q2) -> binary    f binop pes (self q1, self q2)
           Label l pq                -> label     f l (self pq)
           RelExpr pe syms           -> relExpr   f pe syms
-          Antijoin q1 q2            -> antijoin  f (self q1) (self q2)
+          Exists b q1 q2            -> existsf   f b (self q1) (self q2)
         fix g = let x = g x in x
 
 times :: PrimQuery -> PrimQuery -> PrimQuery
@@ -129,6 +129,12 @@ times q q' = Product (q NEL.:| [q']) []
 
 restrict :: HPQ.PrimExpr -> PrimQuery -> PrimQuery
 restrict cond primQ = Product (return primQ) [cond]
+
+exists :: PrimQuery -> PrimQuery -> PrimQuery
+exists a b = Exists True a b
+
+notExists :: PrimQuery -> PrimQuery -> PrimQuery
+notExists a b = Exists False a b
 
 isUnit :: PrimQuery' a -> Bool
 isUnit Unit = True
--- a/src/Opaleye/Internal/Print.hs
+++ b/src/Opaleye/Internal/Print.hs
@@ -9,8 +9,9 @@ import           Opaleye.Internal.Sql (Select(SelectFrom,
                                               SelectJoin,
                                               SelectValues,
                                               SelectBinary,
-                                              SelectLabel),
-                                       From, Join, Values, Binary, Label)
+                                              SelectLabel,
+                                              SelectAntijoin),
+                                       From, Join, Values, Binary, Label, Antijoin)
 
 import qualified Opaleye.Internal.HaskellDB.Sql as HSql
 import qualified Opaleye.Internal.HaskellDB.Sql.Print as HPrint
@@ -23,13 +24,14 @@ import qualified Data.Text          as ST
 type TableAlias = String
 
 ppSql :: Select -> Doc
-ppSql (SelectFrom s)   = ppSelectFrom s
-ppSql (Table table)    = HPrint.ppTable table
-ppSql (RelExpr expr)   = HPrint.ppSqlExpr expr
-ppSql (SelectJoin j)   = ppSelectJoin j
-ppSql (SelectValues v) = ppSelectValues v
-ppSql (SelectBinary v) = ppSelectBinary v
-ppSql (SelectLabel v)  = ppSelectLabel v
+ppSql (SelectFrom s)     = ppSelectFrom s
+ppSql (Table table)      = HPrint.ppTable table
+ppSql (RelExpr expr)     = HPrint.ppSqlExpr expr
+ppSql (SelectJoin j)     = ppSelectJoin j
+ppSql (SelectValues v)   = ppSelectValues v
+ppSql (SelectBinary v)   = ppSelectBinary v
+ppSql (SelectLabel v)    = ppSelectLabel v
+ppSql (SelectAntijoin v) = ppSelectAntijoin v
 
 ppSelectFrom :: From -> Doc
 ppSelectFrom s = text "SELECT"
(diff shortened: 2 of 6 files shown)


src/Opaleye/Internal/Sql.hs
```diff
@@ -108,6 +118,7 @@ baseTable ti columns = SelectFrom $
               , tables = [Table (HSql.SqlTable (PQ.tiSchemaName ti) (PQ.tiTableName ti))] }
 
 product :: NEL.NonEmpty Select -> [HPQ.PrimExpr] -> Select
+product (t NEL.:| []) [] = t
```
tomjaguarpaw: What's going on here?

src/Opaleye/Internal/PrimQuery.hs
```diff
@@ -53,6 +53,7 @@ data PrimQuery' a = Unit
                               (Bindings HPQ.PrimExpr)
                               (PrimQuery' a)
                               (PrimQuery' a)
+                  | Antijoin  (PrimQuery' a) (PrimQuery' a)
```
tomjaguarpaw: Is there a reason this can't be subsumed by the `Join` constructor?

src/Opaleye/Internal/Sql.hs
```diff
@@ -26,6 +26,7 @@ data Select = SelectFrom From
             | SelectValues Values
             | SelectBinary Binary
             | SelectLabel Label
+            | SelectAntijoin Antijoin
```
tomjaguarpaw: Is there a reason this can't be subsumed by the `SelectJoin` constructor?

src/Opaleye/Internal/Sql.hs
tomjaguarpaw: OK, so you're replacing a product of one select by the select itself.

1. Is this necessary to get your patch to work?  If so, why?  If not, let's do this in a different patch.

2. It should probably be done in (or alongside) [`mergeProduct`](https://github.com/ocharles/haskell-opaleye/blob/b070bc23d923097d4cf21d764339e09f846dcb53/src/Opaleye/Internal/Optimize.hs#L25) rather than here

ocharles: This looks like an un-related change, I'll revert that (and try and figure out where it came from!)

src/Opaleye/Internal/PrimQuery.hs
ocharles: It might be possible, but antijoins don't create new bindings so I'm not sure they fit the shape of `Join` - but I'll give it a try tomorrow and see.

src/Opaleye/Internal/Sql.hs
tomjaguarpaw: Looks like a worthy change.  Filing this so we don't lose track of it: https://github.com/tomjaguarpaw/haskell-opaleye/issues/272

src/Opaleye/Internal/PrimQuery.hs
tomjaguarpaw: Oh, now I see that there is no such join type as `ANTIJOIN`, but rather "antijoin" is the name for the pattern "WHERE NOT EXISTS".

No need to stick it in `Join`/`JoinType` then.  It almost certainly won't work.

src/Opaleye/Internal/Print.hs
```diff
@@ -73,6 +75,14 @@ ppSelectLabel l = text "/*" <+> text (defuseComments (Sql.lLabel l)) <+> text "*
                    . ST.replace (ST.pack "*/") (ST.pack " * / ")
                    . ST.pack
 
+ppSelectAntijoin :: Antijoin -> Doc
+ppSelectAntijoin v =
+  text "SELECT *"
+  $$ text "FROM"
+  $$ ppTable (tableAlias 1 (Sql.antijoinTable v))
```
tomjaguarpaw: Is it really the case that the first clause requires a table alias and the second doesn't?  If so that's very strange on the part of PostgreSQL!

tomjaguarpaw:
Now thinking through the overal goal a bit more ...  Can't we have

````haskell
exists :: QueryArr a b -> QueryArr a (Column PGBool)
````

where the return `QueryArr` always has exactly one row?  And that would only need an `Exists` constructor in `PrimQuery`.

**Comment by ocharles** [comment]
Yea, if you see the original issue that was my first plan - but I can't work out how to structure things with `PrimExpr` and `PrimQuery` living in different modules (and now being mutually recursive).

**Comment by ocharles** [comment]
Oh sorry, I've slightly misunderstood your suggestion. I'll have a look at that!

File: src/Opaleye/Internal/Print.hs (hunk 4 above)
Comment (ocharles) [inline src/Opaleye/Internal/Print.hs]: No, `SELECT * FROM x WHERE NOT EXISTS (SELECT * FROM y)` is just fine. This is just lazy copy and pasting to explore a proof of concept :)

File: src/Opaleye/Internal/PrimQuery.hs (hunk 2 above)
Comment (ocharles) [inline src/Opaleye/Internal/PrimQuery.hs]: Well in relational theory an antijoin is a thing:

> The antijoin, written as R ▷ S where R and S are relations, is similar to the semijoin, but the result of an antijoin is only those tuples in R for which there is no tuple in S that is equal on their common attribute names.

`WHERE NOT EXISTS` is a little bit of an abuse of the terminology, but then all joins were originally meant to be on common attributes rather than a general predicate, so that abuse isn't exactly uncommon

File: src/Opaleye/Internal/Sql.hs (hunk 1 above)
Comment (ocharles) [inline src/Opaleye/Internal/Sql.hs]: The change probably comes from a larger branch I'm working on, which is a set of transformations on `From` (and friends) to simplify the generated queries but retain semantics. More on that in the future though.

Comment (tomjaguarpaw) [inline src/Opaleye/Internal/Sql.hs]: Looking forward to that!

**Comment by ocharles** [comment]
@tomjaguarpaw I'm not really sure what the `Exists` combinator would look like. What kind of SQL are you expecting to generate?

**Comment by tomjaguarpaw** [comment]
````SQL
SELECT * FROM (EXISTS (...))
````

**Comment by ocharles** [comment]
EXISTS is an expression so I don't think you can write that.

**Comment by tomjaguarpaw** [comment]
Yeah, I didn't mean that.  I meant

````psql
=> SELECT (EXISTS (SELECT * FROM TABLE1));
 exists
--------
 t
(1 row)
````

**Comment by ocharles** [comment]
Oh right. That might work, I just need to make sure Pg is smart enough to
use an antijoin in the plan still

**Comment by tomjaguarpaw** [comment]
Yes, definitely worth checking that.

**Comment by ocharles** [comment]
OK, I took a stab at doing this with `EXISTS` and the queries I'm generating need `LATERAL` support to work. It could well be the way I'm actually defining the `exists` combinator is wrong though:

```haskell
exists :: O.QueryArr a b -> O.QueryArr a (Column PgBool)
exists q = O.simpleQueryArr $ \(a, tag) ->
  let (_, pq, _) = O.runSimpleQueryArr q a tag
      symbol = O.Symbol "exists" tag
  in (Column (O.AttrExpr symbol), PrimQuery.Exists symbol pq, O.next tag)
```

I'm using it like this:

```haskell
existsExample :: Query (Part Expr)
existsExample = proc _ -> do
  part <- queryTable -< ()
  anotherPartExists
    <- exists (proc p -> do
                 part <- queryTable -< ()
                 where_ -< partId part ==. partId p)
    -< part
  where_ -< not anotherPartExists
  returnA -< part
```

Which generates

```sql
SELECT
          "PID0_1" as "result1_3", "PName1_1" as "result2_3", "Color2_1" as "result3_3", "Weight3_1" as "result4_3",
          "City4_1" as "result5_3"
        FROM
          (SELECT "PID" as "PID0_1", "PName" as "PName1_1", "Color" as "Color2_1", "Weight" as "Weight3_1", "City" as "City4_1"
           FROM "part" as "T1") as "T1",
          (SELECT
             EXISTS (SELECT *
                     FROM
                       (SELECT
                          "PID" as "PID0_2", "PName" as "PName1_2", "Color" as "Color2_2", "Weight" as "Weight3_2",
                          "City" as "City4_2"
                        FROM "part" as "T1") as "T1"
                     WHERE ("PID0_2") = ("PID0_1")) AS exists_3) as "T2"
        WHERE NOT ("exists_3")
```

Notice how `PID0_1` is referenced in the `WHERE` clause for `EXISTS` - but I can't refer to that unless I use a `LATERAL` join, which we don't support in `opaleye` at the moment.

tomjaguarpaw:
I guess `exists` won't work then.  Perhaps we can have `restrictExists` and make it work like `restrict`, that is, add a restricting clause to the `PrimQuery` that's already been built up in our `QueryArr`.  I suspect that would solve the scoping issues.

* https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Operators.hs#L36

* https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/PrimQuery.hs#L126

tomjaguarpaw:
I guess

````haskell
restrictExists :: QueryArr a b -> QueryArr a ()
````

**Comment by ocharles** [comment]
Ok, but I ultimately need `NOT EXISTS` - so maybe `restrictExists :: Exists -> QueryArr a b -> Query a ()` with `data Exists = Exists | NotExists`?

**Comment by tomjaguarpaw** [comment]
Why not just `restrictNotExists` as well?

**Comment by ocharles** [comment]
Your API, your call :) Both work.

**Comment by tomjaguarpaw** [comment]
The tricky part will be implementing them :)

**Comment by ocharles** [comment]
Ok, I believe I have a "final draft" now. I've added public api symbols, and two tests.

File: src/Opaleye/Operators.hs [hunk 5]
```diff
@@ -37,6 +37,18 @@ restrict :: QueryArr (Column T.PGBool) ()
 restrict = QueryArr f where
   f (Column predicate, primQ, t0) = ((), PQ.restrict predicate primQ, t0)
 
+{-| Add a @WHERE EXSITS@ clause to the current query. -}
+exists :: QueryArr a b -> QueryArr a ()
+exists criteria = QueryArr f where
+  f (a, primQ, t0) = ((), PQ.exists primQ existsQ, t0) where
+    (_, existsQ, _) = runSimpleQueryArr criteria (a, t0)
```
tomjaguarpaw: Shouldn't you extract `t1` here and return it from the top level instead of `t0`?

src/Opaleye/Operators.hs
```diff
@@ -37,6 +37,18 @@ restrict :: QueryArr (Column T.PGBool) ()
...
+
+{-| Add a @WHERE EXSITS@ clause to the current query. -}
+notExists :: QueryArr a b -> QueryArr a ()
+notExists criteria = QueryArr f where
+  f (a, primQ, t0) = ((), PQ.notExists primQ existsQ, t0) where
+    (_, existsQ, _) = runSimpleQueryArr criteria (a, t0)
```
tomjaguarpaw: Shouldn't you extract `t1` here and return it from the top level instead of `t0`?

src/Opaleye/Operators.hs
ocharles: Wouldn't that only matter if we were returning columns from `criteria`? The `EXISTS` entirely removes any bindings introduced by `criteria`.

tomjaguarpaw: I see what you mean, although I think it's probably good practice to use `t1` anyway.

tomjaguarpaw: So yes, please change that, just because it makes the code look more obviously correct.

ocharles: Ok, I think I've got that corrected.

tomjaguarpaw:
Looks good.  Just waiting for Travis now.

PR #277, opened 2017-03-02
Add 'exact' ordering method

### LINKED ISSUE #276: Possible addition: `exactOrder`
**Background**: I'm using Opaleye to write queries as a backend for a Haxl data source. Haxl requires me to run a query giving multiple results, and then fill up some `MVars` with the results from the query.

**Problem**: Haxl does *not* order the `MVars` in any specified way, and you cannot depend on an ordering. Haxl gives you a list of MVars. You must always fill the right `MVar` with the right data result, but the order of the list is not specified. As a result, **I need to return a list of results in an arbitrary order, based on the input list**.

For example, if I write some Haxl code like this:

```haskell
k :: GenHaxl () [String]
k = do
  u1 <- getUserNameById 1
  u2 <- getUserNameById 2
  return [u1, u2]
```

I expect the code to give back the name of the first user, and then the name of the second user. **But**, internally, when I fetch from my data source and Haxl calls my backend, **I might get the MVars to fill in a different order than `[u1, u2]`**. In other words: Haxl would give me a list of MVars, with length 2, but the first MVar in the list needs to have the *second* result (`u2`), and the second MVar needs to have the first result (`u1`).

Note that this is an implementation detail: while the above `k` function would, in fact, return the results in the right order you'd expect, it is simply the backend to Haxl that might get the results out of order.

---

More concretely: I want to run a query like this:

```sql
select name from person
  where (id = 2) or (id = 1)
  order by (id=2, id=1) desc;
```

Assuming there are two columns `name :: PGText` and `id :: PGInt4`, with two entries `(1, "Austin")` and `(2, "Tom")`, the above query will actually return them in reverse order (Tom first, then Austin).

The above query can be extended to any general ordering, it doesn't matter; for example, `order by (id=3, id=1, id=2 desc)` will give you the results in the order `[3, 1, 2]`. This is exactly what is needed for Haxl: I need to look over the list, get the parameters for the query, and then order the results according to the list of MVars.

-------

Possible addition:

```haskell
-- | Perform an exact ordering on the result of a query, the
-- ordering determined by the order of the provided @'NL.NonEmpty'@.
exactOrder :: NL.NonEmpty (Column b) -> (a -> Column b) -> Order a
exactOrder xs k = O.Order $ flip fmap k $ \col ->
  let listOrder = O.ListExpr $ flip NL.map xs $ \b ->
        O.BinExpr (O.:==) (O.unColumn col) (O.unColumn b)
  in
  [(O.OrderOp O.OpDesc O.NullsFirst, listOrder)]
```

This roughly translates to the above query, though larger. It has the same style of API as the other order combinators, although it uses a `NonEmpty`, because `ListExpr` requires a `NonEmpty`. Here's an example from a simple database of mine, a live query run with Haxl:

```sql
SELECT "name1_1" as "result1_2"
FROM (SELECT *
      FROM (SELECT *
            FROM (SELECT *
                  FROM (SELECT "id" as "id0_1",
                               "name" as "name1_1",
                               "age" as "age2_1",
                               "address" as "address3_1"
                        FROM "wednesday"."person" as "T1") as "T1"
                  WHERE ((FALSE OR (2 = ("id0_1"))) OR (1 = ("id0_1")))) as "T1"
            ORDER BY (("id0_1") = 2,("id0_1") = 1) DESC NULLS FIRST) as "T1") as "T1"
```

As you can see, Haxl requested I give the results in the order `[2, 1]`.

---

**Notes**: The above query, for me, translates directly to a bitmap scan, though I don't know if this is really significant: the dataset is small (a few dozen rows), and I'm only querying 2-5 IDs at a time right now in a randomly specified order. I don't even have indexes.

The alternative query I've seen floated for the ("I need to order the query results in an arbitrary way") above is this:

```sql
SELECT name FROM person
  LEFT JOIN (VALUES (2,1),(1,2)) as x(id,ordering) on x.id = person.id
  order by x.ordering;
```

This does a clever `LEFT JOIN` over the values: Postgres will turn this into a Hash join. For my queries this is actually slower; but again I'm running over *very* tiny datasets and indexes as I'm learning, so it's entirely possible that the above join will largely win out over this. However, it's also a bit trickier to write and this is literally the first time I've used Opaleye, so, I don't know exactly how you'd like to do this.

thoughtpolice:
Closes #276

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -532,6 +532,15 @@ testOrderBySame :: Test
 testOrderBySame = it "" $ queryShouldReturnSortBy (O.desc fst <> O.asc fst)
                                (flip (Ord.comparing fst) <> Ord.comparing fst)
 
+testOrderExact :: Test
+testOrderExact = it "" $ testH (O.orderBy (O.exact cols snd) table1Q) (result `shouldBe`)
+  where cols   = map O.constant [300,200::Int]
+        result = [ (2::Int, 300::Int)
+                 , (1, 200)
+                 , (1, 100)
+                 , (1, 100)
+                 ]
+
 limitOrderShouldMatch :: (Query (Column O.PGInt4, Column O.PGInt4) -> Query (Column O.PGInt4, Column O.PGInt4))
            -> ([(Int, Int)] -> [(Int, Int)]) -> (PGS.Connection -> Expectation)
 limitOrderShouldMatch olQ ol = testH (olQ (orderQ table1Q))
@@ -985,6 +994,7 @@ main = do
         testOrderBy
         testOrderBy2
         testOrderBySame
+        testOrderExact
       describe "count" $ do
         testCountRows0
         testCountRows3
--- a/src/Opaleye/Internal/Order.hs
+++ b/src/Opaleye/Internal/Order.hs
@@ -1,5 +1,9 @@
 module Opaleye.Internal.Order where
 
+import           Data.Function (on)
+
+import qualified Data.List.NonEmpty as NL
+
 import qualified Opaleye.Column as C
 import qualified Opaleye.Internal.Column as IC
 import qualified Opaleye.Internal.Tag as T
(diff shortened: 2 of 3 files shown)


tomjaguarpaw:
This is great stuff.  A few suggestions:

* It's probably better to use `:=` than `:==` (although now I think about it I don't know why we even have `:==`).

* I suggest you make the first argument `[IC.Column b]` and run `nonEmpty :: [a] -> Maybe (NonEmpty a)` on it.  In the `Nothing` case just return `mempty`.

* Is this technically correct: "in this case, any clauses after the exact ordering are applied only to the results which were not exactly ordered"?  I would assume that clauses after the exact ordering are applied to the exactly ordered rows in the case there is more than one row with any given column value (which can happen if the column we are exactly ordering on is not `UNIQUE`).

* Question: is `HPQ.OrderOp HPQ.OpDesc HPQ.NullsFirst` used, or is it a redundant dummy value that's ignored and only needed because of the structure of the AST?  If the latter, could you insert a comment to that effect?

Otherwise it looks great.  If you address those points I'm happy to merge.

thoughtpolice:
I've pushed an updated commit that addresses 1, 2, and 4. Regarding 4, I left a comment in the code, but briefly: yes, `DESC` is mandatory, and as a result some ordering on the null values is too (because `OrderOp` requires a value for both `OrderDirection` and `OrderNulls`).

I'll double check point 3 soon and get back on it.

tomjaguarpaw:
Oh, and

* There should be a test for this.  Somewhere around [the order tests](https://github.com/tomjaguarpaw/haskell-opaleye/pull/277#issuecomment-283820011) should be good.  Ideally I'd like everything to have property tests but they're rather laborious to write in practice.

tomjaguarpaw:
This is a great PR and I'd like to get it merged ASAP.  Could you add a test? Then I'm happy to merge.

thoughtpolice:
Yeah, sorry about that. I'll clarify point 3, I've been working on something else the past day or so...

thoughtpolice:
And add the test, of course

tomjaguarpaw:
That's no problem.  48 hours to the first PR is a new Opaleye record I
think!

thoughtpolice:
There's now a test, and I also updated the method name: it's now `exact`, as I found `exactOrder` a bit redundant, and I don't find that *too* rude of a name clash. (Well, no more rude than `asc` or `desc` might already be -- but if they are people are qualifying it anyway I imagine).

Still need to update the docs and double check 3 still, sorry. So wait one tiny bit longer.

tomjaguarpaw:
OK, thanks.

thoughtpolice:
I double checked and confirmed you were right; e.g. with the example from #276, running

```
SELECT * from person
  ORDER BY (address='Knowhere'), "age" DESC NULLS FIRST;
```

does indeed use the second clause to tiebreak, as `address` is not `UNIQUE`. I'm still getting used to SQL (I'm a baby) so I should have expected this I suppose. :)

I've simply deleted the offending paragraph from the documentation, since it's otherwise unnecessary (.e.g you would expect `exact` to work just as any other `ORDER BY`, so there is no need to call this out).

Thus, this is ready to merge!

tomjaguarpaw:
This is really great, thanks a lot.  I'm impressed by how quickly you have picked up Opaleye.

PR #279, opened 2017-03-03
Add PGNum/PGFractional/PGString instances for base types

ocharles:
The `rel8` project doesn't use `PGInt` types, preferring instead to use the `Prelude` base types. At the moment, this prevents the use of `Column` as I can't use the `Num` instance due to the `PGNum` constraint. At the moment, I have a custom `Expr` type which is exactly what `Column` is. With these instances, I think I can drop `Expr` entirely.

### DIFF
--- a/src/Opaleye/PGTypes.hs
+++ b/src/Opaleye/PGTypes.hs
@@ -6,7 +6,7 @@
 
 module Opaleye.PGTypes (module Opaleye.PGTypes) where
 
-import           Opaleye.Internal.Column (Column)
+import           Opaleye.Internal.Column (Column, unsafeCoerceColumn)
 import qualified Opaleye.Internal.Column as C
 import qualified Opaleye.Internal.PGTypes as IPT
 
@@ -22,7 +22,7 @@ import qualified Data.ByteString.Lazy as LByteString
 import qualified Data.Time as Time
 import qualified Data.UUID as UUID
 
-import           Data.Int (Int64)
+import           Data.Int (Int8, Int16, Int32, Int64)
 
 import qualified Database.PostgreSQL.Simple.Range as R
 
@@ -38,6 +38,16 @@ instance C.PGNum PGInt8 where
 instance C.PGFractional PGFloat8 where
   pgFromRational = pgDouble . fromRational
 
+instance C.PGNum Int where pgFromInteger    = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Int8 where pgFromInteger   = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Int16 where pgFromInteger  = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Int32 where pgFromInteger  = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Double where pgFromInteger = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Float where pgFromInteger  = unsafeCoerceColumn . pgDouble . realToFrac
+
+instance C.PGFractional Float
+instance C.PGFractional Double
+
 instance C.PGIntegral PGInt2
 instance C.PGIntegral PGInt4
 instance C.PGIntegral PGInt8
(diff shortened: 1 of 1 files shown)


src/Opaleye/PGTypes.hs
```diff
@@ -38,6 +38,16 @@ instance C.PGNum PGInt8 where
 instance C.PGFractional PGFloat8 where
   pgFromRational = pgDouble . fromRational
 
+instance C.PGNum Int where pgFromInteger    = unsafeCoerceColumn . pgInt4 . fromIntegral
```
bergmark: On 64 bit this is would make insertion unsafe. However we decide on this instance an alternative would be to tell people to use Int32/Int64.

src/Opaleye/PGTypes.hs
```diff
@@ -38,6 +38,16 @@ instance C.PGNum PGInt8 where
...
+instance C.PGNum Int8 where pgFromInteger   = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Int16 where pgFromInteger  = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Int32 where pgFromInteger  = unsafeCoerceColumn . pgInt4 . fromIntegral
+instance C.PGNum Double where pgFromInteger = unsafeCoerceColumn . pgInt4 . fromIntegral
```
bergmark: An Int64/pgint8 could be added as well.

bergmark:
Another option would be to add a separate typeclass for "haskell" types. We did it this way in silk-opaleye/girella since we didn't want to use the pg types by accident - or at all. I'm not saying that is a better solution for opaleye itself though, and this change has no effect on our library.

ocharles:
Yea, the definitions do need a bit of tidying up, this was mostly to open up conversation. What would this separate type class look like?

tomjaguarpaw:
I'm pretty sure I'm never going to merge this.  Morally there is a kind `PGType` which `PGInt4`, `PGText` etc. inhabit, and `Column :: PGType -> *`.

tomjaguarpaw:
I suspect your best bet is to use a type family which maps Haskell types to PGTypes and/or vice versa.

ocharles:
That's a shame given that they are perfectly valid instances.

tomjaguarpaw:
They're only valid "by accident".  Morally, `PGNum :: PGType -> Constraint`, not `:: * -> Constraint`.

tomjaguarpaw:
Can you do something like

````haskell
type family Expr a

type instance Expr Int = Column PGInt4
...
````

ocharles:
I'd rather not, I haven't really found any benefit to having separate column types. Ultimately you map to Haskell when you select anyway, so I've found it easier to just move that assumption up front. If these instances can't be added, I'll just stick with my newtype wrapper over column.

ocharles:
I suppose the bigger question is what is having a separate universe of types actually achieving? That is, as you say there is morally a `data PGType = PGInt | PGBool | ..` and `data Column (t :: PGType)` - but I'm not clear what this gains us. Dropping this reduces the surface of the API which has benefits of requiring users hold loss in their head.

tomjaguarpaw:
> I suppose the bigger question is what is having a separate universe of types actually achieving?

Fair question.  My answers would be

1. It correctly reflects reality, which is always advisable.

2. It allows us to have many-to-many mappings between Haskell types and Postgres types.

On the other hand

* Several people have said that they don't like `PGTypes` and no one has ever said they like them.  That doesn't mean that the latter don't exist and it doesn't even mean there are fewer of the latter than the former.  However, it is a good suggestion that there should be a blessed way of mapping between Haskell types and Opaleye types.

ocharles:
I am not convinced 2 is even a good thing - we already have the ability to create new views on types in Haskell - `newtype` wrappers. The moment you open up many-to-many mappings, inference goes out the window unless you already know both sides of the mapping. For 1 you can correctly reflect reality by suitably constraining the type under `Column` with type classes (and indeed that's what types like `FromField` and `ToField` are already doing).

tomjaguarpaw:
I'm not claiming it's a good thing.  I'm claiming it's reality.  There is, a priori, no principled way of converting between Postgres types and Haskell types.  They are simply different programming systems with different types.  Often there's a choice that looks "best", but it's all very ad hoc.  I don't mind (and never have minded) adding an adhoc translation layer as the main API, but one way or another the "true" representation has to be there underneath.

PR #285, opened 2017-03-28
Tutorial: Change the names of groups of columns from 'column' to 'row'

patrickmn:
I'm new to Opaleye so please feel free to shut this down ruthlessly on the grounds of ignorance.

I was a little confused by the terminology in the tutorials in that rows are referred to as columns. Is this actually because they are "functorially" equivalent?

### DIFF
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -126,7 +126,7 @@ synonyms.  For example:
 
 > data Birthday' a b = Birthday { bdName :: a, bdDay :: b }
 > type Birthday = Birthday' String Day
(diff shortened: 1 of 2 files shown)


patrickmn:
Seems like it. Would it be worth adding a note to the documentation explaining this?

tomjaguarpaw:
> Seems like it.

What seems like what?

patrickmn:
It seemed to be a common idiom, and I thought I was missing something obvious.

I agree `Columns` would be clearer too.

I can leave this PR open for discussion, or close it and open an issue.

tomjaguarpaw:
Probably best to close it and open an issue referencing this PR.

PR #287, opened 2017-04-03
Make test database runtime configurable

phadej:


tomjaguarpaw:
Please enter the discussion here: https://github.com/tomjaguarpaw/haskell-opaleye/pull/287

tomjaguarpaw:
Please enter the discussion here: https://github.com/tomjaguarpaw/haskell-opaleye/issues/288

phadej:
you can run tests by either
```
% POSTGRES_CONNSTRING="user=ogre" stack test
```

or 

```
% cat .env 
POSTGRES_CONNSTRING="user='ogre'"
% stack test
```

without envvar set and without `.env` file error is raised:
```
% stack test
...                            
test: user error (Set POSTGRES_CONNSTRING environment variable)
```

cc @k0001 @xldenis

xldenis:
When can this get merged?

tomjaguarpaw:
I'm afraid that my quietness on this PR is because I still haven't fully accepted that envvars are the right way to go.  I still prefer a configuration file.  I'll have to look at the other PR that was pursuing a configuration file approach and compare the two.

phadej:
@tomjaguarpaw this PR supports both, envvars and config file.
- First it tries to use `POSTGRES_CONNSTRING` envvar, but if that fails
- it reads `.env` file where the same variable could be specified
- otherwise fails.

IMHO this is best of both words: you can have your default setup in `.env` but it's easy override by setting the envvar if needed (e.g. travis doesn't need to write config file).

tomjaguarpaw:
OK, I'll have to read through it again because I've been busy and completely forgot where we were up to on this issue!  Sorry about that.

tomjaguarpaw:
OK, ultimately this is fine.  Thanks @phadej!  I'll do a little bit of a refactor on it shortly.

PR #303, opened 2017-06-02
Indexing on PGArray

tstat:
**This is incomplete and I am not proposing merging it as is.**

I would like to open up discussion concerning a couple of considerations in the implementation of `index` for `PGArray` values:

1. In Postgresql arrays are one-based instead of zero-based as they are in Haskell. I assumed the desired behavior would be to have queries be zero-based for Opaleye and have implemented it as such below. If that seems unnatural to you then let me know.

2. The type for `index` is currently `(C.PGNum n, C.PGIntegral n) => Column (T.PGArray a) -> Column n -> Column (C.Nullable a)`. Due to the non-ideal nature of `Nullable` this disallows the indexing of `Column (PGArray (Nullable a))`. It looks like this problem was circumvented in the implementation of the JSON accessor functions by assuming the `PGIsJson` value is `Nullable` and then the user simply coerces the potentially non-nullable type to a nullable type with `toNullable`. Unfortunately, I don't see how this same thing could be accomplished for `index` without the introduction of a function with the signature `Column (PGArray a) -> Column (PGArray (Nullable a))`, or expecting the user to use `unsafeCoerceColumn`, both of which seems pretty clunky to me. What do you think is the correct approach to this problem?


tomjaguarpaw:
1. I'd prefer to stick as close to Postgres semantics as possible.  We'll run into trouble if we try to pretend Opaleye is Haskell that just happens to run on the DB.

2. If you can't extract array elements without converting them to `Nullable` when what's the point of ever using an array with non-nullable elements anyway?

tstat:
1. That makes sense, I will revert the one-based -> zero-based conversion

2. If I am understanding your argument correctly I would say it sounds equivalent to: "What is the point of using `[]` with non-`Maybe` elements when `!!` returns a `Maybe` anyway?". In short, because accessing the array elements by index is not the only thing you intend to do with PG arrays.

tomjaguarpaw:
2. Fair enough

Perhaps this is an opportunity to move closer to where we should be with nullable columns anyway.  I think `Column Nullable a` and `Column NonNullable a` are what Opaleye should support in the future, because `Column (Nullable (Nullable a))` is meaningless, as you have observed.

Therefore, how about we have `PGArray_ Nullable a` and `PGArray_ NonNullable a`, and `type PGArray = PGArray_ NonNullable` to ease the transition?

tstat:
I like this plan, I can make these changes later today.

tomjaguarpaw:
How about we stick with `index`

````haskell
(C.PGNum n, C.PGIntegral n) => Column (T.PGArray a) -> Column n -> Column (C.Nullable a)
````

As you point out, you can get out weird values like `Column (Nullable (Nullable a))`, but perhaps we could have a function `joinNullable :: Column (Nullable (Nullable a))` as a workaround until nullability gets fixed.

**Comment by tstat** [comment]
That works for me

File: src/Opaleye/Internal/HaskellDB/PrimQuery.hs [hunk 1]
```diff
@@ -60,6 +60,7 @@ data BinOp      = (:==) | (:<) | (:<=) | (:>) | (:>=) | (:<>)
                 | (:->) | (:->>) | (:#>) | (:#>>)
                 | (:@>) | (:<@) | (:?) | (:?|) | (:?&)
                 | (:&&) | (:<<) | (:>>) | (:&<) | (:&>) | (:-|-)
+                | OpArrayIndex
```
tomjaguarpaw: I don't understand why you've shoehorned this into `BinOp`.  It seems to me that an additional `ArrayIndex PrimExpr PrimExpr` for `PrimExpr` makes more sense.

tstat: It seemed natural to me following precedent set by `UnOp`, note that `UnOp` is considered one `PrimExpr` although it comes in various syntactical forms (function, prefix, postfix). Each of these forms are then broken down into corresponding `SqlExpr`s in the `defaultSqlExpr` function.

Why should binary operations, in their different syntactical forms (infix, array subscript), follow a different pattern?

tomjaguarpaw: I'm not sure the historical reasons that all `BinOp`s are infix, but since they are it seems neatest to keep it that way for now and add a separate `ArrayIndex` constructor.

tstat: I really must protest once more, I hate to break the logical consistency that I outlined above without good reason. I would argue that it is neater to be consistent in that all unary operators, regardless of syntax, are found under `UnOp` and all binary operators, regardless of syntax, are found under `BinOp`.

If you feel strongly that this is not intuitive or is somehow less organized I will implement it the other way, but then I would suggest following renames:

`BinExpr` -> `InfixBinExpr`
`BinOp` -> `InfixBinOp`

tomjaguarpaw: I definitely want this as a new constructor.  If you want to open a new issue or PR to discuss the organisation of that module then please go ahead.  Such a discussion doesn't belong in this PR though.

tstat: Very well, thank you for hearing me out, I have implemented it as a `PrimExpr` in the latest commit

tomjaguarpaw:
This is great, thanks.

PR #304, opened 2017-06-03
Introduce nullable column type

tstat:
Preparing for the eventual move away from `data Nullable = Nullable a` as discussed in #303.

### DIFF
--- a/src/Opaleye/Internal/Column.hs
+++ b/src/Opaleye/Internal/Column.hs
@@ -1,4 +1,9 @@
+{-# LANGUAGE TypeSynonymInstances
+  , FlexibleInstances
+  #-}
+
 module Opaleye.Internal.Column where
+
 
 import Data.String
 
@@ -10,12 +15,14 @@ import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 --
 -- Do not use the 'Show' instance of 'Column'.  It will be deprecated
 -- in version 0.6.
-newtype Column pgType = Column HPQ.PrimExpr deriving Show
+newtype Column_ colType pgType = Column HPQ.PrimExpr deriving Show
+type Column = Column_ NonNullable
 
 -- | Only used within a 'Column', to indicate that it can be @NULL@.
 -- For example, a 'Column' ('Nullable' @PGText@) can be @NULL@ but a
 -- 'Column' @PGText@ cannot.
 data Nullable a = Nullable
+data NonNullable
 
 unColumn :: Column a -> HPQ.PrimExpr
 unColumn (Column e) = e
--- a/src/Opaleye/Internal/Binary.hs
+++ b/src/Opaleye/Internal/Binary.hs
@@ -1,8 +1,12 @@
-{-# LANGUAGE MultiParamTypeClasses, FlexibleContexts #-}
+{-# LANGUAGE MultiParamTypeClasses
+  , FlexibleContexts
+  , TypeSynonymInstances
+  , FlexibleInstances
+  #-}
 
 module Opaleye.Internal.Binary where
 
-import           Opaleye.Internal.Column (Column(Column))
+import           Opaleye.Internal.Column (Column_(Column), Column)
 import qualified Opaleye.Internal.Tag as T
 import qualified Opaleye.Internal.PackMap as PM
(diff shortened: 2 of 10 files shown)


src/Opaleye/Internal/Column.hs
```diff
@@ -10,12 +15,14 @@ import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 --
 -- Do not use the 'Show' instance of 'Column'.  It will be deprecated
 -- in version 0.6.
-newtype Column pgType = Column HPQ.PrimExpr deriving Show
+data Column_ colType pgType = Column HPQ.PrimExpr deriving Show
```
tomjaguarpaw: Why not `newtype`?

src/Opaleye/Internal/Column.hs
```diff
@@ -10,12 +15,14 @@ import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
...
+type Column = Column_ NonNullable
 
 -- | Only used within a 'Column', to indicate that it can be @NULL@.
 -- For example, a 'Column' ('Nullable' @PGText@) can be @NULL@ but a
 -- 'Column' @PGText@ cannot.
 data Nullable a = Nullable
+data NonNullable
```
tomjaguarpaw: I don't understand how this would work since `Nullable` and `NonNullable` are of different kinds.  Shouldn't we have `data NonNullable a = NonNullable`?

src/Opaleye/Internal/Column.hs
tstat: My mistake, switched back to `newtype`

src/Opaleye/Internal/Column.hs
tstat: Perhaps I misunderstood your outline in #303, but I thought the eventual implementation would be:

```haskell
newtype Column a b = Column HPQ.PrimExpr

data Nullable
data NonNullable
```
Where we then have:  `Column Nullable a` and `Column NonNullable a`

But I cannot switch `Nullable` to kind `*` until we are ready to make the full switch, since it would break everything in the current form.

tomjaguarpaw: Your understanding of the eventual implementation is correct.  But can we really write `Column Nullable b` and `Column NonNullable b` when `Nullable` and `NonNullable` have different kinds?  Perhaps we can if the type parameter `a` is kind polymorphic.

tstat: In fact I don't see how it would work any other way since if `Nullable` is of kind `* -> *` we cannot prevent `Nullable (Nullable a)`

tomjaguarpaw: Yes, eventually `Nullable, NonNullable :: *` (or even `:: Nullability`) but since at the moment `Nullable :: * -> *` it seems to me that we also have to have `NonNullable :: * -> *`.

tstat: No we cannot write `Column Nullable b` and `Column NonNullable b` when `Nullable` and `NonNullable` have different kinds, but I didn't think that is what this intermediate step was about. My understanding was we are simply laying the foundation for a change that will be abrupt in a future release, but you seem to have a different expectation.

Can you clarify your expectation for this step?

I don't understand the value of `NonNullable` and `Nullable` with kind `* -> *` since that would allow even more nonsensical things, such as `Nullable (NonNullable a)`

tomjaguarpaw: Actually I assumed we were just going to touch `PGArray` at this step and leave `Column` alone.  What were you envisaging?

tstat: Ok, my mistake then, I assumed since you wanted to lay the groundwork for `PGArray` having a nullability parameter which would not be in use yet, but would be put to use in a future release, it would make sense to also make sense to lay the groundwork for `Column` to have the nullability parameter, which would also not be in use yet, but would be put to use in the future.

tomjaguarpaw:
OK, let's forget fixing nullability for now and think about `PGArray` again: https://github.com/tomjaguarpaw/haskell-opaleye/pull/303#issuecomment-306000955

PR #305, opened 2017-06-04
Add Nullability parameter to Column

tstat:
Implemented as discussed in #303

### DIFF
--- a/src/Opaleye/Internal/Column.hs
+++ b/src/Opaleye/Internal/Column.hs
@@ -10,7 +10,10 @@ import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
 --
 -- Do not use the 'Show' instance of 'Column'.  It will be deprecated
 -- in version 0.6.
-newtype Column (n :: Nullability) pgType = Column HPQ.PrimExpr deriving Show
+newtype Column' (n :: Nullability) pgType = Column HPQ.PrimExpr deriving Show
+
+type Column = Column' 'NonNullable
+type NullableColumn = Column' 'Nullable
 
 -- | Only used within a 'Column', to indicate that it can be @NULL@.
 -- For example, a 'Column' 'Nullable' @PGText@ can be @NULL@ but a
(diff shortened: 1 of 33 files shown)


src/Opaleye/Internal/Column.hs
```diff
@@ -5,68 +5,70 @@ import Data.String
...
-unsafeEq :: Column a -> Column a -> Column pgBool
+unsafeEq :: Column a b -> Column a b -> Column n pgBool
 unsafeEq = binOp (HPQ.:==)
 
 class PGNum a where
-  pgFromInteger :: Integer -> Column a
+  pgFromInteger :: Integer -> Column 'NonNullable a
 
-instance PGNum a => Num (Column a) where
+instance PGNum a => Num (Column 'NonNullable a) where
```
tomjaguarpaw: It's probably fine to leave this polymorphic over nullability, isn't it?

src/Opaleye/Internal/Operators.hs
```diff
@@ -20,35 +20,38 @@ import qualified Data.Profunctor.Product.Default as D
...
 infix 4 .==
 (.==) :: forall columns. D.Default EqPP columns columns
-      => columns -> columns -> Column T.PGBool
+      => columns -> columns -> Column 'NonNullable T.PGBool
 (.==) = eqExplicit (D.def :: EqPP columns columns)
 
 infixr 3 .&&
 
 -- | Boolean and
-(.&&) :: Column T.PGBool -> Column T.PGBool -> Column T.PGBool
```
tomjaguarpaw: This could probably be polymorphic now.

src/Opaleye/Internal/Column.hs
tomjaguarpaw: Actually, for simplicity, let's leave it how you've got it.

tstat:
There are some operators that I suspect you would want to relax with ambiguous nullability (and some that I would want as well!) but I suspected there would be some amount of discussion over which operators would be best to change, so for this initial commit my intention was to maintain the current type restrictions, and have discussion in this PR for futher changes.

tstat:
@tomjaguarpaw It would be interesting to get your thoughts on this PR. I am not sure how much to build off of this without knowing how you feel about it.

tomjaguarpaw:
The idea is sound[1] and basically The Right Thing™ but it's an *enormous* change and the practical benefit over the status quo is small.  This means that the time before I'll make this change is measured in months and, to be quite realistic, I may never get round to it at all.

I would be very happy for you to take full responsibility for obtaining buy-in from all the major Opaleye stakeholders and implementing the change in the codebase before releasing it as Opaleye 1.0.0.0.  I'm afraid that I can't provide much of my own time into that project though, and you should consider carefully whether it would be a good use of your time.

[1] To be precise, my latest thinking is that we should have

* kinds `PGType` and `PGTypeWithNullability`

* type constructors

    * `Nullable, NonNullable :: PGType -> PGTypeWithNullability`

* a type constructor `Column :: (PGType -> PGTypeWithNullability) -> PGType -> PGTypeWithNullability`

tstat:
Hey Tom, thanks for the reply.

Given that it is extremely common (and desired) that users implement their own `PGType`'s, wouldn't having the kind `PGType` require having open data kinds?

tomjaguarpaw:
Hmm, good question.  Are you saying that open data kinds do exist, or don't exist but would need to?

tstat:
I'm saying that they don't exist but would need to 

https://ghc.haskell.org/trac/ghc/ticket/11080

tomjaguarpaw:
OK.  Maybe it's best to stick with what we've got then.

PR #314, opened 2017-07-31
Provide a cursor interface useful for writing streaming producers

basvandijk:
This is useful for writing streaming producers. See the [following discussion on the haskell-pipes list](https://groups.google.com/forum/#!topic/haskell-pipes/lRtRtibmq9U).

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -37,7 +37,7 @@ library
     , case-insensitive    >= 1.2     && < 1.3
     , bytestring          >= 0.10    && < 0.11
     , contravariant       >= 1.2     && < 1.5
-    , postgresql-simple   >= 0.5     && < 0.6
+    , postgresql-simple   >= 0.5.3   && < 0.6
     , pretty              >= 1.1.1.0 && < 1.2
     , product-profunctors >= 0.6.2   && < 0.9
     , profunctors         >= 4.0     && < 5.3
--- a/src/Opaleye/Internal/RunQuery.hs
+++ b/src/Opaleye/Internal/RunQuery.hs
@@ -4,6 +4,7 @@ module Opaleye.Internal.RunQuery where
 
 import           Control.Applicative (Applicative, pure, (*>), (<*>), liftA2)
 
+import qualified Database.PostgreSQL.Simple.Cursor  as PGSC (Cursor)
 import           Database.PostgreSQL.Simple.Internal (RowParser)
 import           Database.PostgreSQL.Simple.FromField
   (FieldParser, FromField, fromField, pgArrayFieldParser)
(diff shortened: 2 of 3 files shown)


tomjaguarpaw:
This looks great, thanks!

`prepareQuery` only returns `Nothing` when the query is guaranteed to have zero rows. I don't think it's wise to infect the type of `declareCursor` with this internal information.  Could we put knowledge of this into the `Cursor` type?  For example

data Cursor haskells = EmptyCursor | Cursor (FR.RowParser haskells) PGSC.Cursor 

I'm somewhat disinclined to remove `module Opaleye.RunQuery` from the exports because then there's a risk that I forget to export something.  Is there a reason this can't remain, perhaps at the bottom of the export list?

basvandijk:
> prepareQuery only returns Nothing when the query is guaranteed to have zero rows. I don't think it's wise to infect the type of declareCursor with this internal information. Could we put knowledge of this into the Cursor type?

I had this initially but I didn't like that in this case `foldForward` has to perform a case analysis each time it's called. However, I now think might be a premature optimisation so I'll go with your suggestion.

> I'm somewhat disinclined to remove module Opaleye.RunQuery from the exports because then there's a risk that I forget to export something. Is there a reason this can't remain, perhaps at the bottom of the export list?

I think it's better if Cursor is an abstract datatype. This ensures we can modify its implementation later without affecting users. If everything is exported by default Cursor is not abstract anymore.

tomjaguarpaw:
> I had this initially but I didn't like that in this case foldForward has to perform a case analysis each time it's called. However, I now think might be a premature optimisation so I'll go with your suggestion.

Great.  Let me know when you've updated the PR.

> I think it's better if Cursor is an abstract datatype. This ensures we can modify its implementation later without affecting users. If everything is exported by default Cursor is not abstract anymore.

Oh right.  Please define it in `Opaleye.RunQuery.Internal` then, and reexport the type from `Opaleye.RunQuery`.  The policy of Opaleye is that everything should be exported.  Things that are internals and unsupported go in `Internal` modules.

tomjaguarpaw:
By the way, it's also Opaleye policy to support back to GHC 7.6, so you'll need to import `pure` and `<$>`.

basvandijk:
Sure will do!

basvandijk:
What about this?

tomjaguarpaw:
This looks great.  I'm just waiting for the tests to finish.  This should be a formality, since your commit is not changing anything tested!

tomjaguarpaw:
Oops, didn't mean to close.

tomjaguarpaw:
Thanks, that's great!

basvandijk:
Thanks!

When do you expect you'll publish a new release?

tomjaguarpaw:
Just now :) https://hackage.haskell.org/package/opaleye-0.5.4.0

PR #321, opened 2017-08-15
Fix typo

centromere:


tomjaguarpaw:
That's great, thanks.  It looks like the type given in that section is wrong too.  Do you agree?

i.e. https://github.com/tomjaguarpaw/haskell-opaleye/issues/322

centromere:
Yes, good catch.

Issue #341, opened 2017-10-31
when offset and limit are used together the behavior is different depending on the order of the calls


sras:
```
  offset 100 $ limit 100 $ queryTable bookingTable
```

Ends up with this query

```
SELECT "id0_1" as "result1_2",
	       "total_amount1_1" as "result2_2",
	       "pending_amount2_1" as "result3_2",
	       "currency3_1" as "result4_2"
	FROM (SELECT *
	      FROM (SELECT *
	            FROM (SELECT *
	                  FROM (SELECT *
	                        FROM (SELECT *
	                              FROM (SELECT "id" as "id0_1",
	                                           "total_amount" as "total_amount1_1",
	                                           "pending_amount" as "pending_amount2_1",
	                                           "currency" as "currency3_1"
	                                    FROM "booking" as "T1") as "T1") as "T1"
	                        LIMIT 100) as "T1") as "T1"
	            OFFSET 100) as "T1") as "T1"
```

While 

```
limit 100 $ offset 100 $ queryTable bookingTable
```

End up with query

```
SELECT "id0_1" as "result1_2",
	       "total_amount1_1" as "result2_2",
	       "pending_amount2_1" as "result3_2",
	       "currency3_1" as "result4_2"
	FROM (SELECT *
	      FROM (SELECT *
	            FROM (SELECT *
	                  FROM (SELECT *
	                        FROM (SELECT *
	                              FROM (SELECT "id" as "id0_1",
	                                           "total_amount" as "total_amount1_1",
	                                           "pending_amount" as "pending_amount2_1",
	                                           "currency" as "currency3_1"
	                                    FROM "booking" as "T1") as "T1") as "T1"
	                        OFFSET 100) as "T1") as "T1"
	            LIMIT 100) as "T1") as "T1"
```

The OFFSET/LIMIT locations are swapped, which results in the surprising behavior where former query returning zero number of rows, while the latter returning 100 rows as expected.

A raw postgresql query behaves as expected..

```
scratch=# select * from booking offset 10 limit 10;
 id | total_amount | pending_amount | currency
----+--------------+----------------+----------
 11 |           10 |              5 | INR
 12 |           10 |              5 | INR
 13 |           10 |              5 | INR
 14 |           10 |              5 | INR
 15 |           10 |              5 | INR
 16 |           10 |              5 | INR
 17 |           10 |              5 | INR
 18 |           10 |              5 | INR
 19 |           10 |              5 | INR
[...]

PR #343, opened 2017-11-06
Added explanatory docs about how to use limit/offset operators.

### LINKED ISSUE #342: Explain how limit and order interact
@sras expressed confusion about how `limit` and `offset` interact.  https://github.com/tomjaguarpaw/haskell-opaleye/issues/341 It may be surprising for someone who's used to SQLs way of doing things when `OFFSET` and `LIMIT` are attached to the same `SELECT`.

We should explain this in the docs.

saurabhnanda:
Fixes #342 

Wasn't able to get this to build on my machine because of the following error on master:

```
Saurabhs-MBP:haskell-opaleye saurabhnanda$ stack build

Error: While constructing the build plan, the following exceptions were encountered:
       
       In the dependencies for opaleye-0.6.0.0:
           postgresql-simple-0.5.1.1 must match >=0.5.3 && <0.6 (latest applicable is 0.5.3.0)
       needed since opaleye-0.6.0.0 is a build target.
       
Plan construction failed.
```

Hope it works, though.

### DIFF
--- a/src/Opaleye/Order.hs
+++ b/src/Opaleye/Order.hs
@@ -72,6 +72,29 @@ descNullsLast = O.order HPQ.OrderOp { HPQ.orderDirection = HPQ.OpDesc
 {- |
 Limit the results of the given query to the given maximum number of
 items.
+
+/WARNING:/ If you're planning on using limit/offset together please use
+'offset' /before/ you use 'limit'. eg:
+
+@
+limit 10 $ offset 50 yourQuery
+@
+
+This is because of how Opaleye nests queries while applying each operator.
+The result of the query given above is the following, which will return
+10 rows after skipping the first 50 (probably what you want).
+
+@
+SELECT * FROM (SELECT * FROM yourTable OFFSET 50) LIMIT 10
+@
+
+However, reversing the order of the limit\/offset will result in the
+following, which will result in /no rows being returned/ (probably
+not what you want).
+
+@
+SELECT * FROM (SELECT * FROM yourTable LIMIT 10) OFFSET 50
+@
 -}
 limit :: Int -> Query a -> Query a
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
(It looks like you have `postgresql-simple-0.5.1.1` but Opaleye needs `>= 0.5.3`)

tomjaguarpaw:
Thanks!

saurabhnanda:
I have the following stack.yaml, and I haven't edited it: https://github.com/vacationlabs/haskell-opaleye/blob/fix/limit-offset/stack.yaml

tomjaguarpaw:
Is there any particular reason you needed to specify the version explicitly?

saurabhnanda:
> Is there any particular reason you needed to specify the version explicitly?

I didn't. Isn't this the exact stack file committed to the master branch?

tomjaguarpaw:
Interesting.  I don't use Stack so I don't really know about that file.  I've opened a ticket https://github.com/tomjaguarpaw/haskell-opaleye/issues/345 to work out how to deal with it.  Thanks for the report.

src/Opaleye/Order.hs
```diff
@@ -72,13 +72,39 @@ descNullsLast = O.order HPQ.OrderOp { HPQ.orderDirection = HPQ.OpDesc
 {- |
 Limit the results of the given query to the given maximum number of
 items.
+
+/WARNING:/ If you're planning on using limit/offset together please use
+'offset' /before/ you use 'limit'. eg:
+
+@
+limit 10 $ offset 50 yourQuery
```
phadej: should thee be `limitOffset 10 50` operator, to make this common use case more obvious?

tomjaguarpaw: Maybe.  See https://github.com/tomjaguarpaw/haskell-opaleye/issues/341#issuecomment-341110622

PR #346, opened 2017-11-07
Update stack.yaml

mitchellwrosen:


tomjaguarpaw:
Will this pin `product-profunctors` to version `0.7.1.0` for `opaleye` too?  Is it possible to give different `extra-deps` for the different packages?

If not I suggest we just kick `opaleye-sqlite` out of this `stack.yaml`.  It's not up to date or really supported at all.

mitchellwrosen:
> Will this pin product-profunctors to version 0.7.1.0 for opaleye too? Is it possible to give different extra-deps for the different packages?

Yep and nope.

> If not I suggest we just kick opaleye-sqlite out of this stack.yaml. It's not up to date or really supported at all.

Sure.

tomjaguarpaw:
> > Will this pin product-profunctors to version 0.7.1.0 for opaleye too? Is it possible to give different extra-deps for the different packages?

> Yep and nope.

Seems a bit strange.  Perhaps you're supposed to have two different `stack.yaml`s if you want to have different dependencies.

saurabhnanda:
I can confirm that this stack file works, and can be merged.

tomjaguarpaw:
Thanks very much.

PR #360, opened 2017-12-26
Eliminate case expression if all branches are the same either way

duairc:


tomjaguarpaw:
Hello.  Thanks for this.  I'm sorry I haven't responded.  It must have got lost because it was Christmas time.

This sounds like a very nice idea.  I'd probably prefer that the optimization was done on the [`PrimQuery`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/HaskellDB/PrimQuery.hs) rather than the SQL types, though.  It should probably live in [Optimize.hs](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/Optimize.hs).

tomjaguarpaw:
[The offending line seems to be an `ifThenElseMany`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/3ba290e51abb4bcf4b85b777bb2f41b43542ab8b/src/Opaleye/FunctionalJoin.hs#L65).  I imagine you get this behaviour when `fR rr == f lr rr`.  Is that right?

tomjaguarpaw:
> I imagine you get this behaviour when `fR rr == f lr rr`. Is that right?

I've checked, and yes, that's the issue.

I'm a bit surprised that Postgres itself doesn't do this.  Anyway, I'm happy to add the optimization but it should be done on the [`PrimQuery`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/HaskellDB/PrimQuery.hs).

PR #375, opened 2018-03-07
fix for double quoting cast

reygoch:
@tomjaguarpaw  I've implemented the fix recommended in #197 #issuecomment-371183995. My code works now and user is properly inserted.


tomjaguarpaw:
This is an interesting test failure.  The change I suggested writes array types as, e.g., `"float8[]"` when it should be `"float8"[]`. I will think about the best way to address this.  In the meantime feel free to keep using this branch with the change unless you need arrays.

PR #379, opened 2018-03-17
Fix time range queries

### LINKED ISSUE #377: Invalid query from pgRange due to nested casts
I'm writing a query for a table with a `PGRange PGTimestamptz` column, that returns rows whose range encompasses a supplied timestamp. It seems that the `pgRange` function is broken for cases where the provided column-making function itself uses a cast.

Here's a minimal example:

```haskell
data User' a b =
    User { key       :: a
         , sysPeriod :: b
         } deriving (Show, Generic, Eq, Ord)

$(makeAdaptorAndInstance "pUser" ''User')

type User  = User' Int64           (PGSR.PGRange UTCTime)
type UserR = User' (Column PGInt8) (Column (PGRange PGTimestamptz))
type UserW = User' (Column PGInt8) (Maybe (Column (PGRange PGTimestamptz)))

table :: Table UserW UserR
table = Table "users" $ pUser User
    { key       = required "id"
    , sysPeriod = optional "sys_period"
    }

queryAt :: UTCTime -> Query UserR
queryAt t = proc () -> do
    row <- queryTable table -<  ()
    restrict -< sysPeriod row `overlap` pgRange pgUTCTime (Inclusive t) (Inclusive t)
    returnA -< row

```

The generated SQL is:
```sql
       "sys_period1_1" as "result2_2"
FROM (SELECT *
      FROM (SELECT "id" as "id0_1",
                   "sys_period" as "sys_period1_1"
            FROM "users" as "T1") as "T1"
      WHERE (("sys_period1_1") && (CAST('[CAST('2018-03-12T15:04:22.230544Z' AS timestamptz),CAST('2018-03-12T15:04:22.230544Z' AS timestamptz)]' AS tstzrange)))) as "T1"
```
This leads to the error message `syntax error at or near "2018"`.

The query is fixed if I eliminate the inner casts.

BlackBrane:
I found that solving the issue in a satisfactory, non-hacky way required changing the representation in `PrimExpr`, as we need to express a single bound as a function argument, rather than expressing the whole range together. Particularly since `pgRange` is what knows about the name of the range type/constructor functions.

Fixes #377

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -785,7 +785,7 @@ testJsonGetFieldText dataQuery = it "" $ testH q (`shouldBe` expected)
             Arr.returnA -< O.toNullable c1 O..->> O.pgStrictText "c"
         expected :: [Maybe T.Text]
         expected = [Just "21"]
-      
+
 -- Special Test for Github Issue #350 : https://github.com/tomjaguarpaw/haskell-opaleye/issues/350
 testRestrictWithJsonOp :: (O.PGIsJson a) => Query (Column a) -> Test
 testRestrictWithJsonOp dataQuery = it "restricts the rows returned by checking equality with a value extracted using JSON operator" $ testH query (`shouldBe` table8data)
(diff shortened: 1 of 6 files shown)


tomjaguarpaw:
This is great!  Well done for understanding the codebase so easily.

It seems to me that `Inclusive` and `Exclusive` are no longer used.  Could you please get rid of them?

tomjaguarpaw:
... or rather merge them into single constructor called `Bounded` or something.

tomjaguarpaw:
(In `PrimQuery`, not `Range`)

BlackBrane:
Thanks, I'm glad you're satisfied with the patch :)

You make a good point about `BoundExpr` so I've gone ahead fixed it.

tomjaguarpaw:
I used your work to produce a slightly different version that required fewer changes.

https://github.com/tomjaguarpaw/haskell-opaleye/commit/0992b7efa93b38b0ca8364ef515442a2ebb4dddf

Thanks for this!

PR #380, opened 2018-03-25
"liesWithin" predicate function for ranges

BlackBrane:
This adds a function that tests whether a value lies within a range.

Currently this requires checking overlap with a trivial range consisting of two identical inclusive bounds, which isn't the best user experience. For me this is motivated because we're using temporal tables, so we often need to find a record associated with a particular time. Being able to do that without extra ceremony is valuable, and it's nice to get that out of the box.

The name `liesWithin` is my best idea for something that isn't likely to be confused with any other kind of inclusion test.

Its interesting that this is expressed in Postgres the exact same way as the JSON inclusion tests (`<@` and `@>`). It might be nice if the same operators could be used, but the overloading there seems pretty ad hoc, so maybe we shouldn't inherit it.


tomjaguarpaw:
This is great, thanks!

(I also realised that ranges should be easy to property test https://github.com/tomjaguarpaw/haskell-opaleye/issues/381)

PR #382, opened 2018-03-27
Implement Upsert Behavior (Closes #139)

### LINKED ISSUE #139: Support for PostgreSQL INSERT ... ON  CONFLICT statements
As documented at http://www.postgresql.org/docs/9.5/interactive/sql-insert.html#SQL-ON-CONFLICT

AnthonySuper:
In order to provide upsert behavior (as discussed in #139), we made a few simple modifications: First, we add a `Maybe OnConflict` to `SqlInsert`. We then provide an `SqlGenerator` that sets this paramater to `Just DoNothing`. From there we just need to modify the `insert` functions to take an `SqlGenerator`.

The end result is a `runInsertManyIgnoreConflicts` function, which... runs an insert many that ignores conflicts. There's a few variants of other `insertMany` functions as well. 

While this does compile and pass the current test suite, the new functionality is not tested. I wasn't really sure how to properly do so. As a result, it's **probably not the best idea to merge this yet**.

I'm not exactly super experienced in Haskell programming so this implementation might be non-ideal! I'm definitely open to suggestions, especially for how to test it.

### DIFF
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -17,20 +17,23 @@ import qualified Data.ByteString.Char8 as BS8
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.List.NonEmpty as NEL
 
-mkSqlGenerator :: SqlGenerator -> SqlGenerator
-mkSqlGenerator gen = SqlGenerator
+mkSqlGenerator :: SqlGenerator -> Maybe OnConflict -> SqlGenerator
+mkSqlGenerator gen conflict = SqlGenerator
     {
      sqlUpdate      = defaultSqlUpdate      gen,
      sqlDelete      = defaultSqlDelete      gen,
-     sqlInsert      = defaultSqlInsert      gen,
+     sqlInsert      = defaultSqlInsert      gen conflict,
      sqlExpr        = defaultSqlExpr        gen,
      sqlLiteral     = defaultSqlLiteral     gen,
      sqlQuote       = defaultSqlQuote       gen
     }
 
-defaultSqlGenerator :: SqlGenerator
-defaultSqlGenerator = mkSqlGenerator defaultSqlGenerator
 
+defaultSqlGenerator :: SqlGenerator
+defaultSqlGenerator = mkSqlGenerator defaultSqlGenerator Nothing
+
+ignoreConflictsSqlGenerator :: SqlGenerator
+ignoreConflictsSqlGenerator = mkSqlGenerator defaultSqlGenerator (Just DoNothing)
 
 toSqlOrder :: SqlGenerator -> OrderExpr -> (SqlExpr,SqlOrder)
 toSqlOrder gen (OrderExpr o e) =
@@ -60,14 +63,15 @@ defaultSqlUpdate :: SqlGenerator
 defaultSqlUpdate gen tbl criteria assigns
         = SqlUpdate tbl (toSqlAssoc gen assigns) (map (sqlExpr gen) criteria)
 
-
 defaultSqlInsert :: SqlGenerator
+                 -> Maybe OnConflict
                  -> SqlTable
                  -> [Attribute]
                  -> NEL.NonEmpty [PrimExpr]
+                 
                  -> SqlInsert
-defaultSqlInsert gen tbl attrs exprs =
-  SqlInsert tbl (map toSqlColumn attrs) ((fmap . map) (sqlExpr gen) exprs)
+defaultSqlInsert gen conflict tbl attrs exprs =
+  SqlInsert tbl (map toSqlColumn attrs) ((fmap . map) (sqlExpr gen) exprs) conflict
 
 defaultSqlDelete :: SqlGenerator
                  -> SqlTable
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -68,7 +68,6 @@ defaultSqlInsert :: SqlGenerator
                  -> SqlTable
                  -> [Attribute]
                  -> NEL.NonEmpty [PrimExpr]
-                 
                  -> SqlInsert
 defaultSqlInsert gen conflict tbl attrs exprs =
   SqlInsert tbl (map toSqlColumn attrs) ((fmap . map) (sqlExpr gen) exprs) conflict
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -144,12 +144,12 @@ defaultSqlExpr gen expr =
       CastExpr typ e1 -> CastSqlExpr typ (sqlExpr gen e1)
       DefaultInsertExpr -> DefaultSqlExpr
       ArrayExpr es -> ArraySqlExpr (map (sqlExpr gen) es)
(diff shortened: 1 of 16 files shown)


tomjaguarpaw:
Thanks for this!  I haven't had a chance to look over this in detail yet but one thing that stands out is this line:

     sqlInsert      = defaultSqlInsert      gen conflict

It feels like `conflict` should not be passed in as an argument here.  Can you have a look at rearchitecting your patch so that's not required?

src/Opaleye/Internal/HaskellDB/Sql/Print.hs
```diff
@@ -97,11 +97,14 @@ ppDelete (SqlDelete table criteria) =
...
 
 ppInsert :: SqlInsert -> Doc
-ppInsert (SqlInsert table names values)
+ppInsert (SqlInsert table names values onConflict)
     = text "INSERT INTO" <+> ppTable table
       <+> parens (commaV ppColumn names)
       $$ text "VALUES" <+> commaV (parens . commaV ppSqlExpr)
                                   (NEL.toList values)
+      <+> text (case onConflict of
+        Just _ -> "ON CONFLICT DO NOTHING"
```
tomjaguarpaw: Please avoid a wildcard pattern here and use `DoNothing`.  The downside of using a wildcard pattern is that we will not get a compiler warning if we forget to add a future clause for `OnConflict ...`.

AnthonySuper:
I refactored out the wildcard pattern, but I sadly don't see a super clean way to remove passing the conflict parameter as an argument. This is almost certainly due to inexperience, of course.

One way to do it would be to make a separate `mkSqlGenerator` function that takes the `OnConflict` argument, but that would involve duplicating all of the other fields. I bet there's other ways to do it but I'm not clever enough to figure them out, unfortunately.

src/Opaleye/Manipulation.hs
```diff
@@ -61,14 +79,18 @@ runInsertMany :: PGS.Connection
...
+runInsertManyIgnoreConflicts :: PGS.Connection
+                             -> T.Table columns columns'
+                             -> [columns]
+                             -> IO Int64
+runInsertManyIgnoreConflicts = runInsertManyGenerator SD.defaultSqlGenerator
 
 -- | Insert rows into a table and return a function of the inserted rows
 --
--- @runInsertManyReturning@'s use of the 'D.Default' typeclass means that the
+-- @runInsertManyReturning@'s use of the 'D.Default' typeclaarrangeInsertManyss means that the
```
tomjaguarpaw: Is this a find/replace accident?

src/Opaleye/Manipulation.hs
```diff
@@ -257,27 +279,38 @@ arrangeInsert :: T.Table columns a -> columns -> HSql.SqlInsert
 arrangeInsert t c = arrangeInsertMany t (return c)
 
 {-# DEPRECATED arrangeInsertSql
-    "You probably want 'runInsertMany' instead. \
+    "You probably want 'runInsertMany' instead. arrangeInsertMany\
```
tomjaguarpaw: Is this a find/replace accident?

tomjaguarpaw:
I spotted a couple of places where `arrangeInsertMany` popped up unexpectedly.  Was that a find/replace accident or similar?

tomjaguarpaw:
I think you can solve the conflict parameter argument situation if you change the type signature of the `sqlInsert` field here

https://github.com/AnthonySuper/haskell-opaleye/blob/17db649a32918444de4750a5edf89c6285563194/src/Opaleye/Internal/HaskellDB/Sql/Generate.hs#L16

to

    sqlInsert :: SqlTable -> [Attribute] -> NEL.NonEmpty [PrimExpr] -> SqlInsert

Then the rest should fall into place.  Feel free to give this a go today otherwise I hope to get round to it tomorrow.

tomjaguarpaw:
Oops, I mean to

```haskell
sqlInsert :: SqlTable -> [Attribute] -> NEL.NonEmpty [PrimExpr] -> Maybe OnConflict -> SqlInsert
```

tomjaguarpaw:
Can you fetch my latest repo and merge in 96d74d56fb6f89bb16125c4ade0dd3efe37f20b8 to your PR?  I managed to get rid of the argument.  Then we'll give it another check and proceed to merge.

tomjaguarpaw:
In fact, please merge `origin/upsert` (https://github.com/tomjaguarpaw/haskell-opaleye/tree/upsert)

tomjaguarpaw:
No problem about the find/replace errors.  That's what a second pair of eyes are for.

Git seems to have bodged the merge somehow, but that's fine, I can take it from here.  Thanks a lot for your PR!

kahlil29:
Hey @tomjaguarpaw 
any idea when this would be merged?

tomjaguarpaw:
I'm a bit reluctant to merge this without tests, and I haven't got round to writing those yet.  Is it critical for you to have a release of this rather than just something in a branch you can experiment with?  You can use the `origin/upsert` branch now if you want to play around, but I can't guarantee the same API in the final release.

Let me know what your needs are and I'll think of the best way to help you.

src/Opaleye/Internal/HaskellDB/Sql/Default.hs
```diff
@@ -17,20 +17,23 @@ import qualified Data.ByteString.Char8 as BS8
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.List.NonEmpty as NEL
 
-mkSqlGenerator :: SqlGenerator -> SqlGenerator
-mkSqlGenerator gen = SqlGenerator
+mkSqlGenerator :: SqlGenerator -> Maybe OnConflict -> SqlGenerator
+mkSqlGenerator gen conflict = SqlGenerator
```
dfithian: This seems weird to me since it's baking in the behavior of the `on conflict` strategy in to SQL generation. It seems like a SQL generator should generate for as many cases as possible, and I can imagine cases in which I would want to fail on conflicts in one table but not another. So I think this part of the design should be bubbled up farther ([reference](https://github.com/tomjaguarpaw/haskell-opaleye/pull/384/files#diff-170a1d73c68b44c6ced1062ece0f2b92R16)). Does that make sense?

That's really the only thing here I see that's different, but I really appreciate the work being done here!

dfithian:
Regarding tests, I noticed the `addons` section in .travis.yml has postgres version 9.3, but the behavior is not documented until 9.5. Is there alternate data regarding this behavior being present in 9.3?

src/Opaleye/Internal/HaskellDB/Sql/Default.hs
tomjaguarpaw: Your instincts are right.  This has been addressed in https://github.com/tomjaguarpaw/haskell-opaleye/tree/upsert which is the latest version of this PR.

tomjaguarpaw:
Migrating this to https://github.com/tomjaguarpaw/haskell-opaleye/pull/385 as it has changed somewhat from the original version.

tomjaguarpaw:
@dfithian Do you mean `ON CONFLICT`?  I've no idea when that was added to Postgres but if it's not until 9.5 we will indeed have to bump the Postgres version!  Thanks for spotting.

dfithian:
Okay cool I'll write some tests and see if they work in Travis. Locally I
have 9.6

tomjaguarpaw:
@kahlil29 I'd love you to give feedback on `origin/upsert` if you have the time.  Then I'm happy to merge into master and do a release if you want.

kahlil29:
@tomjaguarpaw completely missed this in my notifications, sorry! Just saw this.

Issue #230, opened 2016-11-05
PGNumeric support


lippling:
I only saw a `PGNumeric` instance for `IsSqlType`. Am I missing something or is the PGNumeric support incomplete?

tomjaguarpaw:
There's `instance PGOrd T.PGNumeric` here: https://github.com/tomjaguarpaw/haskell-opaleye/blob/7b5e780513ba77d51f43aa78899a78d87fa93ac5/src/Opaleye/Order.hs#L79

But I guess you are saying that we are missing

````haskell
pgNumeric :: Integer -> Column PGFloat8

instance C.PGNum PGNumeric where
  pgFromInteger = pgNumeric
````

Do we need anything else?  Please feel free to submit a PR!

**Comment by mzabani** [comment]
I'm pretty new to Haskell and even newer to Opaleye, so sorry for sticking in like this. Shouldn't *Numeric* be arbitrary precision decimal numbers? Considering that, wouldn't it be better to use `Rational` in Haskell as the equivalent type for postgresql's *Numeric*? I was thinking something like

`pgNumeric :: (Integral a) => Ratio a -> Column PGnumeric`

https://www.postgresql.org/docs/9.2/static/datatype-numeric.html

**Comment by tomjaguarpaw** [comment]
Sure, that sounds better.

**Comment by saurabhnanda** [comment]
Just got bitten by this. Is the `Default` instance absent because there's no real consensus on *what* PGNumeric should map to on the Haskell side?

**Comment by tomjaguarpaw** [comment]
It's absent because no one's ever got round to doing anything about it.  According to the postgresql-simple documentation `PGNumeric` can map to `Scientific` and `Ratio Integer`.  Does anyone want to submit a PR?

https://hackage.haskell.org/package/postgresql-simple-0.5.0.0/docs/Database-PostgreSQL-Simple-FromField.html#v:pgArrayFieldParser

**Comment by saurabhnanda** [comment]
Will submit a PR after getting it to work in our app.

**Comment by saurabhnanda** [comment]
So, we need **both** instances, right? `Default` AND `QueryRunnerColumnDefault`?

**Comment by saurabhnanda** [comment]
which [`HPQ.Literal`](https://www.stackage.org/haddock/lts-7.14/opaleye-0.5.2.2/src/Opaleye.Internal.HaskellDB.PrimQuery.html#Literal) should I be using to complete the following:

```
[...]

PR #383, opened 2018-04-03
Numeric Support

blinky3713:
Attempt to address #230 

It's difficult to read through the confusion in #230 to figure out what the highlighted concerns were, but I'm just following the lead from `postgresql-simple` here and using the `Scientific` type. My actual use case is  storing 32-byte unsigned integers, which I'm able to do now with tests passing using the following newtype in my application code.

```haskell
newtype Value =
  Value { unValue :: Integer }
    deriving (Eq, Show, Enum, Num, Ord, Real, Integral)

instance QueryRunnerColumnDefault PGNumeric Value where
  queryRunnerColumnDefault = Value . truncate . toRational <$> fieldQueryRunnerColumn @Scientific

instance Default Constant Value (Column PGNumeric) where
  def = Constant $ constant @Scientific . fromInteger . unValue
```

I understand that "unbounded" integers are probably a common usecase, so if you want I can put them in this pr as well.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -38,6 +38,7 @@ library
     , pretty              >= 1.1.1.0 && < 1.2
     , product-profunctors >= 0.6.2   && < 0.10
     , profunctors         >= 4.0     && < 5.3
+    , scientific          >= 0.3     && < 0.4
     , semigroups          >= 0.13    && < 0.19
     , text                >= 0.11    && < 1.3
     , transformers        >= 0.3     && < 0.6
--- a/src/Opaleye/Constant.hs
+++ b/src/Opaleye/Constant.hs
@@ -13,6 +13,7 @@ import qualified Data.Text                       as ST
 import qualified Data.Text.Lazy                  as LT
 import qualified Data.ByteString                 as SBS
 import qualified Data.ByteString.Lazy            as LBS
+import qualified Data.Scientific                 as Sci
 import qualified Data.Time                       as Time
 import qualified Data.UUID                       as UUID
 
@@ -72,6 +73,9 @@ instance D.Default Constant ST.Text (Column T.PGText) where
 
 instance D.Default Constant LT.Text (Column T.PGText) where
   def = Constant T.pgLazyText
+
+instance D.Default Constant Sci.Scientific (Column T.PGNumeric) where
+  def = Constant T.pgNumeric
 
 instance D.Default Constant Int (Column T.PGInt4) where
   def = Constant T.pgInt4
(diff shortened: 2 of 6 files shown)


blinky3713:
Its unclear to me that my changes are causing this test failure, will wait to hear back before investigating

tomjaguarpaw:
The build failure was transient.

This is great!  Can you squash the patches and improve the commit message?  Then I'll merge.

blinky3713:
Sure, but did you want me to go ahead implement a `PGBigInt` type with an `Integer` representation?

tomjaguarpaw:
Isn't the `PGNumeric` you've implemented sufficient for that?

blinky3713:
I would like to add

```haskell
data PGBigInt

pgBigInt :: Integer -> Column PGBigInt
pgBigInt = IPT.literalColumn . HPQ.NumericLit . fromInteger

instance QueryRunnerColumnDefault T.PGBigInt Integer where
  queryRunnerColumnDefault = truncate . toRational <$> fieldQueryRunnerColumn @Sci.Scientific

```

etc to the library

tomjaguarpaw:
But `bigint` is an eight byte (64 bit) integer: https://www.postgresql.org/docs/current/static/datatype.html

So why not use `Int64` instead of `Integer`?

blinky3713:
Sorry, maybe `BigInt` is a bad name, but basically what I'm saying is that I have a need to go beyond 64 bit Integers -- I need 32 bytes in my case. 

`PGNumeric` is a bad representation (because I know they are integers, and I don't want to have to cast every time), in theory it could make for unsafe selects or joins. This is why I'm proposing a `PGInteger` or `PGBiggerInt` or whatever we want to call it, with the caveat that you might fail at runtime by trying to cast an `Integer` larger than is supported by `numeric` (unlikely because `numeric` is ungodly large.  

If you don't agree, it's fine because I will just continue to use the newtype in my application.

tomjaguarpaw:
Since your "big int" isn't actually a Postgres standard type it belongs in a client library, not Opaleye itself.

tomjaguarpaw:
I don't see the new commit messages.

blinky3713:
I had amended the commit messages, thinking you would squash-merge. Now I have squashed the two commits with a single amended message, and resubmitted.

tomjaguarpaw:
Ah, sorry I had a thinko.  This looks great.  I'll merge when I the test has completed.

tomjaguarpaw:
Great stuff.  Thanks.

PR #384, opened 2018-04-05
Conflict resolution for insertions

dfithian:
Went for it! #139


tomjaguarpaw:
Thanks!

Actually there is already [a PR for this](https://github.com/tomjaguarpaw/haskell-opaleye/pull/382).  The latest version is my `upsert` branch.

Could you compare and contrast your approach with @AnthonySuper's and mine?

https://github.com/tomjaguarpaw/haskell-opaleye/compare/master...upsert

I think they're basically the same but it would be good to get another pair of eyes on it.

Issue #244, opened 2016-11-29
Support for `SELECT DISTINCT ON`


duairc:
At the moment is there any support for PostgreSQL's [DISTINCT ON](https://www.postgresql.org/docs/current/static/sql-select.html#SQL-DISTINCT) feature?

If not (I can't see it anywhere), what would be the best work-around in the mean time?

tomjaguarpaw:
Currently there's no support for `DISTINCT ON`.

Could you aggregate into arrays and then take the first element of each array?

fieldstrength:
I have also come to a situation where this `distinct on` support would be very useful to me.

I'm curious @tomjaguarpaw if you have any thoughts on an implementation plan that would be acceptable? This is valuable enough that I might probably give it a go, if its not too heavy a lift.

Some observations and questions:

* According to the [Postgres documentation](https://www.postgresql.org/docs/current/static/sql-select.html#SQL-DISTINCT) the allowed expressions are similar to those of `ORDER BY`, so the implementation might look similar. Like orders, we could use the nice semigroup interface, where different field projections can be appended together. So I'm thinking a reasonable interface would be something like

```haskell
data DistinctOn a 
-- = DistinctOn (a -> NonEmpty PrimExpr)

instance Semigroup DistinctOn
instance Contravariant DistinctOn

distinguish :: (a -> Column b) -> DistinctOn a

distinctOn :: DistinctOn a -> Query a -> Query a
```
* What should be done about our `distinct` function, currently implemented as a `groupBy` aggregation? While it might be nice to handle it together with this functionality, or actually make the implementation match the name, it seems like a better first step to consider this as an independent feature.

* There is a potential gotcha in that if one simultaneously uses `DISTINCT ON` with `ORDER BY`, then any columns appearing in the former have to appear before (to the left) of any other columns in the latter. While this was at first a worry, I believe this pitfall is sidestepped if we ensure that this new feature is kept at a different level of nesting as the ordering. If I have understood correctly, this is what the `simpleQueryArr` function does, is that correct?

tomjaguarpaw:
Hi @fieldstrength.  Yes, please give this a go!  I'd probably just go for

```haskell
data DistinctOnPP a b = ...

instance ProductProfunctor DistinctOnPP where
    ....

distinctOn :: Default DistinctOnPP b b => (a -> b) -> Query a -> Query a
```

I think it's probably sufficient for `type DistinctOnPP = Unpackspec`, actually.

[...]

PR #392, opened 2018-06-21
Support DISTINCT ON

fieldstrength:
Resolving #244 

I went with your basic idea of using `Default Unpackspec` as a way to get at the expressions to distinguish. That seems to work nicely! However, the most straightforward implementation turned out not to work due to the interaction of this feature with ordering.

In order for my use case (and most of them I would guess) we need a way to determine which row is selected. It turns out we can't just accomplish this by ordering first, at a deeper nesting. A `DISTINCT ON` query does not respect any ordering except the one provided with it directly. Therefore we need a construct that can perform both of these modifications together.

With that in mind, here are some properties and remarks about what I've implemented currently.

### Exposed API

```haskell
distinctOn   :: Default Unpackspec b b => (a -> b)            -> Select a -> Select a
distinctOnBy :: Default Unpackspec b b => (a -> b) -> Order a -> Select a -> Select a
```

The first function is exactly what we discussed. The second allows us to control the ordering used to choose representative rows from each equivalence class. It differs from the raw SQL in that it does not allow a user to input an invalid query, because any expressions that are distinguished are automatically included as the leftmost ordering expressions. (Obviously this makes no difference, its just fixing a leaking implementation detail)

_Alternative:_ The second function could be called `distinctOnOrderBy` but I fear this would trick someone into believing the _output_ was ordered!

### Internal changes

The main opinionated property here is that I used a single constructor of `PrimQuery` that can represent both ordering and/or `distinct on`. Any constructor that could represent the needed functionality can also represent both sub-features, so I saw no need to keep a duplicate. I also chose a representation that makes bad SQL unrepresentable, namely by interpreting all of the `distinct on` expressions as leftmost `order by _ asc` expressions too. One could argue this implicitness is undesirable, but I think it is less so than allowing to represent invalid queries, and I cannot think of any way this would cause any real issues.

### DIFF
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -45,7 +45,11 @@ data PrimQuery' a = Unit
                   | Product   (NEL.NonEmpty (PrimQuery' a)) [HPQ.PrimExpr]
                   | Aggregate (Bindings (Maybe (HPQ.AggrOp, [HPQ.OrderExpr], HPQ.AggrDistinct), HPQ.PrimExpr))
                               (PrimQuery' a)
-                  | Order     [HPQ.OrderExpr] (PrimQuery' a)
+                  -- | Represents both @DISTINCT ON@ and @ORDER BY@ clauses. In order to represent valid
+                  --   SQL only, @DISTINCT ON@ expressions are always interpreted as the first @ORDER BY@s
+                  --   when present, preceding any in the provided list.
+                  --   See 'Opaleye.Internal.Sql.distinctOnOrderBy'.
+                  | DistinctOnOrderBy (Maybe (NEL.NonEmpty HPQ.PrimExpr)) [HPQ.OrderExpr] (PrimQuery' a)
                   | Limit     LimitOp (PrimQuery' a)
                   | Join      JoinType
                               HPQ.PrimExpr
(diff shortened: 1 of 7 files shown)


tomjaguarpaw:
Thanks, this looks very nice!  Could you please explain in more detail what you mean about invalid SQL and "leftmost ordering expressions"?  Can you give an example of an invalid query that could be generated if you didn't take these steps and how your approach prevents that?

fieldstrength:
Sure!

So here's a basic example of something where I might want to use this feature to choose which rows I get:
```sql
select * from (values ('a',1),('a',2),('b',1),('b',2)) as "V";
 column1 | column2
---------+---------
 a       |       1
 a       |       2
 b       |       1
 b       |       2
```

I might choose to use the exposed function like 

```haskell
distinctOnBy fst (asc snd) $ values [...]
```

without this measure, this would generate something which looks like 
```sql
select distinct on (column1) * 
    from (values ('a',1),('a',2),('b',1),('b',2)) as "V" 
    order by column2;
ERROR:  SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: select distinct on (column1) * from (values ('a',1),('a',2),...
```
With this measure we instead get:
```sql
select distinct on (column1) * 
    from (values ('a',1),('a',2),('b',1),('b',2)) as "V" 
    order by column1, column2;
 column1 | column2
---------+---------
 a       |       1
 b       |       1
(2 rows)
```

tomjaguarpaw:
That's interesting.  So it's a syntactic requirement of the SQL.  (Do you understand why?  Maybe for efficiency `DISTINCT ON` is implemented with a sort and this restriction is a way of making it explicit that the sort is  happening.)

https://www.postgresql.org/docs/10/static/sql-select.html#SQL-DISTINCT

src/Opaleye/Internal/Sql.hs
```diff
@@ -153,10 +154,19 @@ aggregate aggrs s = SelectFrom $ newSelect { attrs = SelectAttrs
 aggrExpr :: Maybe (HPQ.AggrOp, [HPQ.OrderExpr], HPQ.AggrDistinct) -> HPQ.PrimExpr -> HPQ.PrimExpr
 aggrExpr = maybe id (\(op, ord, distinct) e -> HPQ.AggrExpr distinct op e ord)
 
-order :: [HPQ.OrderExpr] -> Select -> Select
-order oes s = SelectFrom $
-    newSelect { tables = [s]
-              , orderBy = map (SD.toSqlOrder SD.defaultSqlGenerator) oes }
+distinguishAndOrder :: Maybe (NEL.NonEmpty HPQ.PrimExpr) -> [HPQ.OrderExpr] -> Select -> Select
```
tomjaguarpaw: Can you rename this `distinctOnOrderBy` for consistency?

fieldstrength:
Yeah, I suspect your suggestion is probably right. The ordering Ive seen running these queries seems to suggest that the distinct on is implemented with a sort. And I suppose it might allow them to avoid some annoying special case on their end to simply require the ordering they use be entered by the user directly. Not sure though.

src/Opaleye/Internal/Sql.hs
fieldstrength: Absolutely.

src/Opaleye/Internal/PrimQuery.hs
```diff
@@ -45,7 +45,11 @@ data PrimQuery' a = Unit
                   | Product   (NEL.NonEmpty (PrimQuery' a)) [HPQ.PrimExpr]
                   | Aggregate (Bindings (Maybe (HPQ.AggrOp, [HPQ.OrderExpr], HPQ.AggrDistinct), HPQ.PrimExpr))
                               (PrimQuery' a)
-                  | Order     [HPQ.OrderExpr] (PrimQuery' a)
+                  -- | Represents both @DISTINCT ON@ and @ORDER BY@ clauses. In order to represent valid
+                  --   SQL only, @DISTINCT ON@ expressions are always interpreted as the first @ORDER BY@s
+                  --   when present, preceding any in the provided list.
+                  --   See 'Opaleye.Internal.Sql.distinguishAndOrder'.
```
tomjaguarpaw: And this one :)

tomjaguarpaw:
If you can clean up the git history by fixupping the last two commits (bikeshed and consistency rename) into  the commit they morally belong to then that would be great.  Otherwise I'll do it myself when I get round to it.

fieldstrength:
Sure, I can do this. Would you be opposed to me just making it just one commit? I'm not sure any of the current ones provide much value individually. They were more for my own development purposes. (And they include the earlier idea of how this was going to work)

tomjaguarpaw:
Please keep the two import formatting commits separate and squash everything else.

fieldstrength:
Good idea

tomjaguarpaw:
Thanks!  This is a great patch.  You have clearly understood the internals of Opaleye very well.  Thanks for helping to make Opaleye a better library :)

fieldstrength:
Thank you! I'm very glad to help. I like the library a lot and use it at work.

I feel I'm more or less just starting to get the hang of the internals, so I look forward to the next contribution :)

tomjaguarpaw:
> I look forward to the next contribution :)

Me too!

PR #393, opened 2018-07-01
Range bound accessors, and fix unbounded ranges

fieldstrength:
This was mainly going to be about adding the two accessor functions for range bounds. They seems like the last "obviously needed" pieces to have a nice range interface.  We've been using `lowerBound` in our repo for a while now.

But when adding the test I noticed that we've had the wrong representation for unbounded ranges. So `infinity` is not actually valid for all ranges, its just a value that inhabits certain types which have ranges. 

Its an easy fix so I just included it in this.

From the [documentation](https://www.postgresql.org/docs/current/static/rangetypes.html): 

> Also, some element types have a notion of "infinity", but that is just another value so far as the range type mechanisms are concerned. For example, in timestamp ranges, [today,] means the same thing as [today,). But [today,infinity] means something different from [today,infinity) — the latter excludes the special timestamp value infinity.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1025,6 +1025,22 @@ testRangeAdjacency = it "generates adjacency" $ testH q (`shouldBe` [True])
         range a b = O.pgRange O.pgInt4 (R.Inclusive a) (R.Exclusive b)
         q = A.pure $ (range 1 2) O..-|- (range 2 3)
 
+testRangeBoundsEnum :: forall a b.
+    ( Show a, Eq a, Enum a, O.IsRangeType b
+    , O.QueryRunnerColumnDefault (Nullable b) (Maybe a))
+        => String -> (a -> Column b) -> a -> a -> Test
+testRangeBoundsEnum msg mkCol x y = it msg $ \conn -> do
+    -- bound functions for discrete range types return fields as from the form [x,y)
+    let ranges = [ O.pgRange mkCol (R.Inclusive x) R.PosInfinity
+                 , O.pgRange mkCol R.NegInfinity   (R.Inclusive y)
+                 , O.pgRange mkCol (R.Exclusive x) (R.Exclusive y)
+                 ]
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
This looks really good.  We currently use [postgresql-simple's data type for range bounds](https://hackage.haskell.org/package/pg-0.5.2.1/docs/Database-PostgreSQL-Simple-Range.html#t:RangeBound).  From what you say it seems like it must be buggy because it can't distinguish `infinity)` from `infinity]`.  Do you agree?

tomjaguarpaw:
This is great, thanks.  Maybe in the future we'll need to allow ranges with inclusive infinity but we should probably send that request upstream to postgresql-simple.

fieldstrength:
It seems to me that the problem is actually that `RangeBound` has one too many constructors. It should just have something called `Unbounded` instead of `Pos`/`NegInfinity`, since that's the only concept that the ranges themselves know about. I would support trying to get this fixed in `postgresql-simple` because currently it can express absurd things, like having negative infinity as an upper bound. It would also make it less likely that anyone else would get confused about how ranges work in postgres, as I have.

To support inclusive-infinite bounds. We'd need a way to construct ranges with a different signature than the current `pgRange`. The issue is actually independent of ranges, since `infinity`/`-infinity` are valid date/time values to Postgres, but are not expressible in the Haskell types `Day` or `UTCTime`, and thus not in Opaleye.

It seems to me that the solution is to give a way to provide the infinite values for these types, either via a type class or by defining them as constants, and then provide a way to construct ranges by providing `RangeBound (Colum _)`s directly. That could then replace the current `pgRange`.

tomjaguarpaw:
Do you want to [file an issue on postgresql-simple](https://github.com/lpsmith/postgresql-simple/issues/new)?

`class SqlHasInfinity a where infinity :: a; minusInfinity :: a` seems like a reasonable idea.  What would happen if you tried to read one of those back into Haskell though?

PR #407, opened 2018-11-15
Add MINIMAL pragma to IsSqlType class

zyla:
Both methods have a default implementation, but to be correct one of
them must be implemented.


tomjaguarpaw:
Thanks that's great

PR #411, opened 2018-11-21
Add missing deps to stack.yaml

zyla:


tomjaguarpaw:
Thanks for this.  Can you explain what it is needed for?

tomjaguarpaw:
I don't use stack so I have no idea.

zyla:
The test suites require some dependencies outside of Stackage. So while `stack build` succeeds, `stack test` (before this PR) fails with the message:

```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for opaleye-0.6.7003.0:
    aeson-1.3.1.1 from stack configuration does not match >=0.6 && <1.3  (latest matching version is 1.2.4.0)
    multiset needed, but the stack configuration has no specified version  (latest matching version is 0.3.4.1)
needed since opaleye is a build target.

Some different approaches to resolving this:

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some working build configuration. This can be convenient when dealing with many
    complicated constraint errors, but results may be unpredictable.

  * Recommended action: try adding the following to your extra-deps in /home/zyla/w/o/haskell-opaleye/stack.yaml:

- aeson-1.2.4.0
- multiset-0.3.4.1

Plan construction failed.
```

The fix is to specify versions of the needed dependencies.

Note: while in the above message `stack` suggests adding only `aeson` and `multiset`, after doing this it fails again suggesting the `base-compat` dependency.

zyla:
By default stack only uses packages specified in the "resolver", to guarantee reproducibility. To pull something from Hackage, you need to specify the exact version - again, for reproducibility.

tomjaguarpaw:
I see.  Thanks for explaining and fixing this for Stack users.

Issue #410, opened 2018-11-21
`singletonArray` fails when used at `PGInt8` type


zyla:
The query expression `pure (O.singletonArray (O.pgInt8 1))` generates the following SQL:

```
SELECT
array_prepend(1,CAST(ARRAY[] AS bigint[])) as "result1_1"
FROM (SELECT 0) as "T1"
```

It is rejected by Postgres with the error message `function array_prepend(integer, bigint[]) does not exist`.
Seems that the literal `1` is interpreted as an `integer`.

The solution could be adding a cast somewhere, but I don't know the best place to do it.

Here's a full test case that reproduces the problem:

```diff
diff --git a/Test/Test.hs b/Test/Test.hs
index 53c37b5..da6dcda 100644
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -872,6 +872,11 @@ testArrayIndexOOB = it "returns Nothing when the index is out of bounds" $
   testH (A.pure $ O.pgArray O.pgInt4 [5,6,7] `O.index` O.pgInt4 8)
         (`shouldBe` ([Nothing] :: [Maybe Int]))
 
+testSingletonArray :: Test
+testSingletonArray = it "constructs a singleton PGInt8 array" $
+  testH (A.pure $ O.singletonArray (O.pgInt8 1))
+        (`shouldBe` ([[1]] :: [[Int64]]))
+
 type JsonTest a = SpecWith (Query (Column a) -> PGS.Connection -> Expectation)
 -- Test opaleye's equivalent of c1->'c'
 testJsonGetFieldValue :: (O.SqlIsJson a, O.QueryRunnerColumnDefault a Json.Value) => Query (Column a) -> Test
@@ -1145,6 +1150,7 @@ main = do
         testFloatArray
         testArrayIndex
         testArrayIndexOOB
+        testSingletonArray
       describe "joins" $ do
         testLeftJoin
         testLeftJoinNullable
```

On my machine fails with:

```
  Test/Test.hs:876:22: 
  1) arrays constructs a singleton PGInt8 array
       uncaught exception: SqlError
       SqlError {sqlState = "42883", sqlExecStatus = FatalError, sqlErrorMsg = "function array_prepend(integer, bigint[]) does not exist", sqlErrorDetail = "", sqlErrorHint = "No function matches the given name and argument types. You might need to add explicit type casts."}
```

tomjaguarpaw:
[...]

PR #412, opened 2018-11-29
Cast all literals

### LINKED ISSUE #253: Explicitly cast literals?
Because of issues like https://github.com/tomjaguarpaw/haskell-opaleye/issues/250#issuecomment-265146334 should we explicitly cast literals on creation?  It will make the generated SQL more verbose but more explicit and perhaps ultimately clearer.

zyla:
This patch makes Opaleye add an explicit `CAST` to SQL literals. This is to force Postgres to assign the type we want to the expression.

Not doing this caused problems in the past, as Postgres's automatic type conversion did not always work as expected (and we shouldn't rely on it anyway, since we have all type information statically).

Fixes #253, #410 

Notes:
- I had to change `Opaleye.PGTypes.literalColumn` to have the original definition, because the new `literalColumn` requires an additional constraint.
- `sqlUUID` had to be adjusted, because it relied on `pgString`, generating a cast to `text` instead of the expected `uuid`

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -886,6 +886,11 @@ testArrayIndexOOB = it "returns Nothing when the index is out of bounds" $
   testH (A.pure $ O.pgArray O.pgInt4 [5,6,7] `O.index` O.pgInt4 8)
         (`shouldBe` ([Nothing] :: [Maybe Int]))
 
+testSingletonArray :: Test
+testSingletonArray = it "constructs a singleton PGInt8 array" $
+  testH (A.pure $ O.singletonArray (O.pgInt8 1))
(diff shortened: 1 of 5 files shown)


tomjaguarpaw:
That looks good.  Could you explain why `instance IsSqlType a => IsSqlType (C.Nullable a)` has been removed?

zyla:
It was moved together with `IsSqlType` to avoid an orphan instance (`Nullable` is defined in a different module): https://github.com/tomjaguarpaw/haskell-opaleye/pull/412/files#diff-4c7091beb1ec291d7f9a45d1dc20b5e7R48 .

tomjaguarpaw:
Ah yes, I see.

This is a great patch.  Thanks!

PR #415, opened 2018-12-25
Add `arrayAppend`

zyla:


tomjaguarpaw:
Thanks Macek, that's great!

PR #428, opened 2019-06-05
Type Family style tables template haskell

wraithm:
This adds the ability to have type family style data types easily with some template haskell magic.

This works on data types with a single type argument, and each field uses a `TableField` type, eg.

```haskell
data UserP f = UserP
    { pName        :: TableField f Name SqlText NN Req
    , pUserEmail   :: TableField f Email SqlText NN Req
    , pEnabled     :: TableField f Bool SqlBool NN Opt
    }
```

You can then call:
```haskell
makeTableAdaptorInstanceMap "pUser" ''UserP
```

and have access to `pUser` adaptor, the appropriate `Default` instances, and `Map` type family instances.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -41,6 +41,7 @@ library
     , scientific          >= 0.3     && < 0.4
     , semigroups          >= 0.13    && < 0.20
     , text                >= 0.11    && < 1.3
+    , template-haskell    >= 2.10    && < 2.15
     , transformers        >= 0.3     && < 0.6
     , time                >= 1.4     && < 1.10
     , time-locale-compat  >= 0.1     && < 0.2
@@ -92,6 +93,7 @@ library
                    Opaleye.Internal.TableMaker,
                    Opaleye.Internal.Tag,
                    Opaleye.Internal.TypeFamilies,
+                   Opaleye.Internal.TypeFamiliesTH,
                    Opaleye.Internal.Unpackspec,
                    Opaleye.Internal.Values
                    Opaleye.Internal.HaskellDB.PrimQuery,
--- a/src/Opaleye/Internal/TypeFamiliesTH.hs
+++ b/src/Opaleye/Internal/TypeFamiliesTH.hs
@@ -1,0 +1,177 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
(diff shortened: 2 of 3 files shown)


tomjaguarpaw:
This is very cool.  I've wanted this for some time.  Can we somehow make it more polymorphic in the type family?  Maybe that's too much to hope for ...

wraithm:
Hmm. This probably can be made more generic, but I've had trouble making it more generic in the past. What do you mean by more polymorphic in the type family?

tomjaguarpaw:
Well, we don't just have `TableField` but also `RecordField`.  It would be
nice if it worked for both.

wraithm:
Oh interesting! I don't think I've ever used `RecordField`. Could you give me an example of using it? It's a little hard to understand from just looking at hackage.

tomjaguarpaw:
Oh, it's just the same as `TableField` except for making records that don't necessarily come from tables.  Therefore it doesn't need information regarding whether the field is required or optional.

wraithm:
What's an example of a record that doesn't come from a table? Are these like views or some other projection?

tomjaguarpaw:
It can be any intermediate or final state of your query.  You proposed the record type for the user table

```haskell
data UserP f = UserP
    { pName        :: TableField f Name SqlText NN Req
    , pUserEmail   :: TableField f Email SqlText NN Req
    , pEnabled     :: TableField f Bool SqlBool NN Opt
    }
```

Perhaps you want to return each user with either the email address or a string saying that the user is disabled then you'd want to do something like this


```haskell
data ResultP f = Result
    { rName    :: RecordField f Name SqlText NN
    , rMessage :: RecordField f String SqlText NN
    }

result = fmap (\u -> ResultP { rName = pName u
                             , rMessage =
                                 ifThenElse (pEnabled u)
                                            ("Your email address is " .++ pUserEmail u)
                                            "Your account is disabled" })
```

`ResultP` is not the type of any table but it's a useful thing to carry `Field`s around in nonetheless.

wraithm:
@tomjaguarpaw I added some `RecordField` stuff and a really simple test... If it compiles, then it should work. Idk if there's a better way to test these examples, but this generated the right stuff for your `ResultP` example afaict.

wraithm:
Huh... The pull request isn't updating... I pushed more commits...

wraithm:
Well, I have no idea what's going on with Github. So, please look at my `type-families-th` branch in my fork.

wraithm:
Yeah, I have no idea what's up... I made a new pull request: https://github.com/tomjaguarpaw/haskell-opaleye/pull/430

wraithm:
Yeah, I pushed to my other PR, and it updated that PR. I didn't do anything different... Who knows...

PR #429, opened 2019-06-05
Added postgres-embedded for tests

tomjaguarpaw:


tomjaguarpaw:
@ilya-murzinov This is very cool but is x64 only so I can't run it on my system :(

ocharles:
Have a look at tmp-postgres - that's what we use at work

benkolera:
@ilya-murzinov a bit of feedback: be very very careful downloading unexpected code onto other people's machines. Especially via a non secure link that could be MitM.

https://github.com/ilya-murzinov/postgres-embedded/blob/master/src/Database/PostgreSQL/Embedded/Download.hs#L48

tomjaguarpaw:
I'm generally much happier with tmp-postgres, although it doesn't work on old GHCs which I'd like to keep supporting, so I won't rewrite the tests to use it just yet.

PR #430, opened 2019-06-06
Type Family style template haskell

wraithm:
This adds the ability to have type family style data types easily with some template haskell magic.

This works on data types with a single type argument, and each field uses a TableField type, eg.

```haskell
data UserP f = UserP
    { pName        :: TableField f Name SqlText NN Req
    , pUserEmail   :: TableField f Email SqlText NN Req
    , pEnabled     :: TableField f Bool SqlBool NN Opt
    }
```
You can then call:

```haskell
makeTableAdaptorInstanceMap "pUser" ''UserP
```

and have access to pUser adaptor, the appropriate Default instances, and Map type family instances.

### DIFF
--- a/Test/TypeFamilyTH.hs
+++ b/Test/TypeFamilyTH.hs
@@ -1,15 +1,15 @@
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE TemplateHaskell       #-}
-{-# LANGUAGE TypeFamilies          #-}
-{-# LANGUAGE UndecidableInstances  #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
 
-module Main where
+module TypeFamilyTH where
 
-import           Opaleye.Internal.TypeFamilies
-import           Opaleye.Internal.TypeFamiliesTH
-import           Opaleye.SqlTypes
+import Opaleye.Internal.TypeFamilies
+import Opaleye.Internal.TypeFamiliesTH
+import Opaleye.SqlTypes
 
 
 data UserP f = UserP
@@ -27,8 +27,3 @@ data ResultP f = Result
     }
 
 makeTableAdaptorInstanceMap "pResult" ''ResultP
-
-
--- If this compiles, this works
-main :: IO ()
-main = return ()
--- a/Test/TypeFamilyTH.hs
+++ b/Test/TypeFamilyTH.hs
@@ -1,0 +1,34 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+module Main where
+
+import           Opaleye.Internal.TypeFamilies
+import           Opaleye.Internal.TypeFamiliesTH
+import           Opaleye.SqlTypes
+
+
+data UserP f = UserP
+    { pName      :: TableField f String SqlText NN Req
+    , pUserEmail :: TableField f String SqlText NN Req
+    , pEnabled   :: TableField f Bool   SqlBool NN Opt
+    }
+
+makeTableAdaptorInstanceMap "pUser" ''UserP
+
+
+data ResultP f = Result
+    { rName    :: RecordField f String SqlText NN
+    , rMessage :: RecordField f String SqlText NN
+    }
+
+makeTableAdaptorInstanceMap "pResult" ''ResultP
+
+
+-- If this compiles, this works
+main :: IO ()
+main = return ()
--- a/src/Opaleye/Internal/TypeFamiliesTH.hs
+++ b/src/Opaleye/Internal/TypeFamiliesTH.hs
@@ -1,0 +1,177 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Opaleye.Internal.TypeFamiliesTH (makeTableAdaptorInstanceMap) where
+
+import Data.List (foldl', nub)
+import Data.Maybe (listToMaybe, mapMaybe)
+import Data.Profunctor.Product (ProductProfunctor)
+import Data.Profunctor.Product.Default (Default (..))
+import Data.Profunctor.Product.Internal.TH (adaptorDefinition, adaptorDefinitionFields)
+import Language.Haskell.TH
+import Opaleye.Internal.TypeFamilies ((:<$>), (:<*>), F, IMap, TableField)
+import Opaleye.Map (Map)
+
+
+type DefaultTable h o n r p a b = Default p (TableField a h o n r) (TableField b h o n r)
+
+
+-- Taken, slightly modified, from product-profunctors makeAdaptorAndInstance
+data ConTysFields
+    = ConTys   [Type]         -- ^^ The type of each constructor field
+    | FieldTys [(Name, Type)] -- ^^ The fieldname and type of each constructor field
+    deriving Show
+
+
+fieldTypes :: ConTysFields -> [Type]
+fieldTypes (ConTys ts)   = ts
+fieldTypes (FieldTys xs) = map snd xs
+
+
+data DataDecInfo = DataDecInfo
+    { dTyName  :: Name
+    , dTyVars  :: [Name]
+    , dConName :: Name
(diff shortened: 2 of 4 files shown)


Test/TypeFamilyTH.hs
```diff
@@ -0,0 +1,34 @@
...
+
+data ResultP f = Result
+    { rName    :: RecordField f String SqlText NN
+    , rMessage :: RecordField f String SqlText NN
+    }
+
+makeTableAdaptorInstanceMap "pResult" ''ResultP
+
+
+-- If this compiles, this works
```
tomjaguarpaw: There should probably be a test that the things created have the right type.

src/Opaleye/Internal/TypeFamiliesTH.hs
```diff
@@ -0,0 +1,184 @@
+{-# LANGUAGE ConstraintKinds #-}
```
tomjaguarpaw: Probably slightly better if this module is `Opaleye.Internal.TypeFamilies.TH`.

tomjaguarpaw: It ought to be exposed as `Opaleye.TypeFamilies.TH`, too.

Test/TypeFamilyTH.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+module Main where
+
+import           Opaleye.Internal.TypeFamilies
```
tomjaguarpaw: Is it possible to use `Opaleye.TypeFamilies` instead, i.e. not use the `Internal` module?  That would be preferable.

tomjaguarpaw:
No need to act on these changes whilst I'm considering what to do, unless you particularly want to.

src/Opaleye/Internal/TypeFamiliesTH.hs
tomjaguarpaw: Oh I see, it's exposed in `Opaleye.TypeFamilies`.

src/Opaleye/Internal/TypeFamiliesTH.hs
```diff
@@ -0,0 +1,184 @@
...
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Opaleye.Internal.TypeFamiliesTH (makeTableAdaptorInstanceMap) where
+
+import Data.List (foldl', nub)
+import Data.Maybe (listToMaybe, mapMaybe)
+import Data.Profunctor.Product (ProductProfunctor)
+import Data.Profunctor.Product.Default (Default (..))
+import Data.Profunctor.Product.Internal.TH (adaptorDefinition, adaptorDefinitionFields)
```
tomjaguarpaw: This can't depend on stuff in an internal module from a different package.  If it really needs these I'll have to expose them from product-profunctors.

tomjaguarpaw:
> No need to act on these changes whilst I'm considering what to do, unless you particularly want to.

Actually, I'm working on this now.

tomjaguarpaw:
The test is not actually being compiled!  If I ensure it is compiled with the below patch then I get the error below.  @wraithm could you please have a look?

```
Test/TypeFamilyTH.hs:21:1: error:
    Could not find TableField or RecordField
```

```diff
diff --git a/Test/TypeFamilyTH.hs b/Test/TypeFamilyTH.hs
index 53b6f06..52c870d 100644
--- a/Test/TypeFamilyTH.hs
+++ b/Test/TypeFamilyTH.hs
@@ -5,7 +5,7 @@
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE UndecidableInstances  #-}

-module Main where
+module TypeFamilyTH where

 import           Opaleye.Internal.TypeFamilies
 import           Opaleye.Internal.TypeFamiliesTH
diff --git a/opaleye.cabal b/opaleye.cabal
index 3e5d0c8..590def7 100644
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -108,7 +108,8 @@ test-suite test
   type: exitcode-stdio-1.0
   main-is: Test.hs
   other-modules: QuickCheck,
-                 TypeFamilies
+                 TypeFamilies,
+                 TypeFamilyTH
   hs-source-dirs: Test
   build-depends:
     aeson >= 0.6 && < 1.5,
```

src/Opaleye/Internal/TypeFamiliesTH.hs
wraithm: @tomjaguarpaw I made a PR to `product-profunctors`

tomjaguarpaw: Cool, thanks!

tomjaguarpaw:
Sorry that I'm a bit slow to look at this.  I need to get a chunk of free time to sit down and look it through clearly.

tomjaguarpaw:
I've been mulling this over in a background thread. For some reason I'm hesitant to merge this into Opaleye.  I'm trying to understand why.  I think the reason is that this functionality (in a generic form) really belongs in `product-profunctors`.  I'm uncomfortable adding to the public API of Opaleye something that doesn't really belong and that is not necessarily stable.

Let me present two other options.  Please let me know how you feel about them and how they fit in with your workflow.  I don't want to impede you using this cool new functionality!

1. Easiest, but ultimately we'll migrate away from it to 2: Add this in a separate package (`opaleye-tf-th` or something ...) that depends on `opaleye` and `product-profunctors`.  It can be tied to a specific version of `product-profunctors` so we can use `Internal` modules.

2. A bit more work but makes this "HKD" style stuff accessible to everyone for whatever purpose they like: Make this generic and add it to product-profunctors

tomjaguarpaw:
Your TH is better written than mine so I'd like to merge some of your changes into `product-profunctors` anyway.  I think we should be ambitious and aim for 2.  Either I can do it when I get round to it or you can have a go.  It's something I think I'll get to reasonably quickly (~ 25% chance within the week and 75% chance within the month).

wraithm:
Yeah, that's super interesting. You're totally right that this should all be in `product-profunctors`. It's conceptually generic.

The only two things I worry about are the class constraints on the `Default` instances and that `Map` type family:
- In this PR, `defaultTableConstraint` is dependent on `TableField`/`RecordField`, and I don't know if it's possible to make it fully generic.
- `Map` belongs to `opaleye`. Is it possible to put `Map` into `product-profunctors` too? Would that make sense?

wrt `Default` class constraints, I think I understand conceptually how to make it generic, but it's kind of complicated. You'd need to collect the type variables on the data type you want to make a `Default` instance over, and find all of those variables in the various constructor fields. Maybe it would make sense to have the Template Haskell code take the type variables as an input? I might not be making much sense here... What are your thoughts?

wraithm:
Yeah, I think it's somewhat ambiguous if you're making a `Default` instance for a datatype that has non-product-profunctor-y arguments. So, the new `product-profunctor` template-haskell should have two versions:
- The function tries to infer the type variables with some simple heuristics. ie. all of the type variables are profunctor-y
- The function takes in an explicit list of type variables that are profunctor-y

Here's an example of the explicit case:

Say that `roles` is just simply a type variable for having the data type be polymorphic, but `roles` isn't a type variable for a product-profunctor. You'd need to mark `f` as the special case. `makeTableAdaptorInstanceMapExplicit` would be a template haskell function for marking `f` explicitly for the `Default` instance.

```haskell
data User f roles = User
    { userName :: TableField f Name SqlText NN Req
    , userRoles :: Roles f roles
    }

makeTableAdaptorInstanceMapExplicit "pUser" 'User ["f"]
```

wraithm:
Maybe another heuristic to use would be to look for parameters of type-level arity 2 ie. Kind signatures of `* -> * -> *`. That won't capture every case, hence the need for the explicit type parameter TemplateHaskell code, but that's maybe going to do more automatically than just assuming that all of the type parameters are product-profunctor-y. Though, using this heuristic, I think you could end up with some false positives, and get some erroneous code gen.

Thoughts?

tomjaguarpaw:
Hmm, if we're making `pUser` then we know how to generate code for the `TableField ...` field.  For the `Roles ...` field don't we just use `pRoles`?

wraithm:
Yep, we just use pRoles for the adaptor definition. Though, the question about marking which type variables are product-profunctor-y as opposed to just templated things is more about the `Default` instances and the type signature for the adaptor. You need the explicit marking for figuring out what needs to be mapped over vs what needs to be simply passed through.

But, that brings up a really good point... I won't necessarily be able to find the adaptor needed to make a nested data type... How will I know if `pRoles` is defined and named? Do I need to make an adaptor in a where clause for every nested data type field? Hm... That sounds like the right solution rather than trying to find existing adaptors, just generate ones on the fly.

wraithm:
Here's an example for the adaptor and instance output that I'm imagining for `pUser`:

```haskell
pUser (UserP n rs) = UserP ***$ lmap userName n **** lmap userRoles (pRoles rs)
  where pRoles = ... -- Defined here locally because it'd be difficult to figure out how to find pRoles if it already existed

instance 
( ProductProfunctor p
, Default p (TableField a Name SqlText NN Req) (TableField b Name SqlText NN Req)
{- insert all of the Default constraints needed for all the fields of Roles that use the f parameter here -}
=> Default p (User a roles) (User b roles) where
  def = pUser (UserP def (Roles def def def ...))
```

tomjaguarpaw:
Hmm, I'm beginning to think this is significantly more complex than I anticipated.  When I get some time to sit down and think it through I will do so.  Please let me know what more you discover in the mean time.

Issue #435, opened 2019-06-30
Query building performance discussion


tomjaguarpaw:
Migrating from https://github.com/tomjaguarpaw/haskell-opaleye/issues/434

tomjaguarpaw:
I would be very happy to have benchmarks in the CI.  That would be a good way of catching performance regressions early.  I'm also open to improving rendering performance by switching away from `String`.  What exactly did you have in mind?  I'm surprised that the `pretty` package is not already optimised as much as possible!

csasarak:
I had a look at the `pretty`[ source code](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/src/Text.PrettyPrint.HughesPJ.html#render) and it appears that while the `Doc` type is a rope structure, when it comes to rendering it uses `Strings`  which have the O(n) cost of appending. I'm actually not too surprised, the documentation says its well suited to things like compilers and if the whole point is to generate user-facing messages then the slowness probably isn't worth fixing. That SPJ did a lot of work on it suggests to me also that this is the case. 

I was running into some trouble with the more complex queries in my program at work (I do data warehousing/analytics), what I found is that if I dump the generated code into postgres it runs quite quickly, but for whatever reason opaleye was taking a long time. When I profiled, most of the time was attributed to `formatAndShowSql` and the only CAF that it creates is the query `String`. You can see an example of flamegraph (made with `ghc-prof-flamegraph`) showing this [here.](https://imgur.com/a/rCB7VSk) This graph shows where time is spent proportionally, but for larger queries it can turn into a lot of real time. 

For most of my simpler queries this isn't much of a problem, but I have a few which aggregate data across multiple tables, with multiple left-joins, inner joins, and unions. Opaleye is really great because I can abstract a lot of the common behavior into query combinators, but maybe this is abusive?

In any case, you can see a really rough non-working example of what I'm envisioning here: https://github.com/csasarak/haskell-opaleye/tree/bytestrings

The bytestring builders have O(1) append and `postgresql-simple` appears to be able to handle them directly. I can try to work on an example in the benchmark repo that I worked in for the `unionAll` question. I'd like to convince someone else that this is indeed a problem or that it can be done better before pushing too hard for change.

tomjaguarpaw:
> Opaleye is really great because I can abstract a lot of the common behavior into query combinators, but maybe this is abusive?

This is exactly what Opaleye is designed for so this is definitely a use case I want to perform well.

[...]

PR #436, opened 2019-07-01
Test using the prettyprinter library for rendering SQL

csasarak:
This is a test based on the discussion in #435. My benchmarks (on a relatively simple query) show it to be a bit faster than the old one. I want to try using the `fuse` function from `prettyprinter` to see if that can help since many of the printing functions add a few relatively small things to the append. 

It's pretty rough, and if this is a direction we're interested in going I'll need some guidance on design.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,11 @@ env:
 - GHCVER=8.2.2
 - GHCVER=8.4.4
 - GHCVER=8.6.5
+- GHCVER=8.8.1
 - GHCVER=head
 
 script:
-- curl -sSL https://raw.github.com/tomjaguarpaw/neil/stable5/travis.sh | sh -s tomjaguarpaw stable5
+- curl -sSL https://raw.github.com/tomjaguarpaw/neil/4ab6ea4/travis.sh | sh -s tomjaguarpaw 4ab6ea4
 
 # To clear the caches go to
 #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.6.7004.1
+
+* Fixed quadratic slowdown in `removeEmpty`.
+
+* Fixed `read` compatibility with time-1.9 in test suite.
+
 ## 0.6.7004.0
 
 * Many changes to the documentation that use the new names.  See entry
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -43,19 +43,19 @@ there is no support for creating databases or tables, though these
 features may be added later according to demand.
 
 A table is defined with the `table` function.  The syntax is
-simple.  You specify the types of the columns, the name of the table
-and the names of the columns in the underlying database, and whether
-the columns are required or optional.
+simple.  You specify the types of the fields, the name of the table
+and the names of the fields in the underlying database, and whether
+the fields are required or optional.
 
 (Note: This simple syntax is supported by an extra combinator that
-describes the shape of the container that you are storing the columns
+describes the shape of the container that you are storing the fields
 in.  In the first example we are using a tuple of size 3 and the
 combinator is called `p3`.  We'll see examples of others later.)
 
 The `Table` type constructor has two arguments.  The first one tells
-us what columns we can write to the table and the second what columns
+us what fields we can write to the table and the second what fields
 we can read from the table.  In this document we will always make all
-columns required, so the write and read types will be the same.  All
+fields required, so the write and read types will be the same.  All
 `Table` types will have the same type argument repeated twice.  In the
 manipulation tutorial you can see an example of when they might differ.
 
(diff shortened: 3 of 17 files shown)


csasarak:
Actually, for fun I tried using this in my real application and it's significantly quicker. I will try to think of some better benchmarks than what I was testing with before.

tomjaguarpaw:
Can you explicitly import `<>` from `Data.Monoid` so it works with old GHCs?

> if this is a direction we're interested in going I'll need some guidance on design.

This definitely looks like the right approach.  I'm happy to provide guidance. What would you like guidance on?

I would suggest a few `type` synonyms and helper functions so as to reduce diff churn, for example

* `type Doc = Data.Text.Prettyprint.Doc SqlStatement`
* `empty = mempty`
* `text = fromString`

Basically let's try to get the diff to be as small as possible.

csasarak:
@tomjaguarpaw I meant exactly what you've done, thanks! I want to make sure this change doesn't break things so I tried to limit the text stuff to its own set of functions. 

One question I have is why `Opaleye.Internal.HaskellDB.Sql.Print` and `Opaleye.Internal.Print` are separate? It looks like one is only for Selects and the other has Update/Delete/Insert. I ask because I'd like to make an explicit export list to help GHC inline a bit better, which might be improved by combining the two modules so that less needs to be exposed. 

I will incorporate your suggestions, thank you!

tomjaguarpaw:
The two modules are separate because the `Internal.HaskellDB` one was lifted directly from the haskelldb package and the `Internal` one was written fresh.  I'm not against merging them if it improves performance.

tomjaguarpaw:
You can change the types from `String` to `Text` if you like.  I can't imagine that that would help performance much though, because they're very small.

One thing we need to consider is what is actually the correct type of thing to create in the first place.  [`runQueryExplicit`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/fde61aa2ee63ae9c1d40dcc50e58e996a9d15d25/src/Opaleye/RunQuery.hs#L69-L74) uses [postgresql-simple's `queryWith_`](https://hackage.haskell.org/package/postgresql-simple-0.6.2/docs/Database-PostgreSQL-Simple.html#v:queryWith).  The latter uses [`Query`](https://hackage.haskell.org/package/postgresql-simple-0.6.2/docs/src/Database.PostgreSQL.Simple.Types.html#Query) which is just a newtyped `ByteString`.  Ideally then we would be generating ByteStrings directly.  Unfortunately it doesn't seem that prettyprinter supports that.  We'd have to generate `Text` and then UTF8-encode the `Text`.

csasarak:
I may give that a try next then.

 I have reduced the diff size and pushed, though some things I kept changed into forms that match the way `prettyprinter` works a bit better.

tomjaguarpaw:
It would still be good to see `empty = mempty`.

tomjaguarpaw:
In fact if you can `import Prelude hiding (++)` and define `(++) = (<>)` then I think we're doing even better.  Ideally we would have zero changes to the body of the module at all.

csasarak:
> You can change the types from `String` to `Text` if you like. I can't imagine that that would help performance much though, because they're very small.

You're right here, probably not much gain to be had.

> One thing we need to consider is what is actually the correct type of thing to create in the first place. [`runQueryExplicit`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/fde61aa2ee63ae9c1d40dcc50e58e996a9d15d25/src/Opaleye/RunQuery.hs#L69-L74) uses [postgresql-simple's `queryWith_`](https://hackage.haskell.org/package/postgresql-simple-0.6.2/docs/Database-PostgreSQL-Simple.html#v:queryWith). The latter uses [`Query`](https://hackage.haskell.org/package/postgresql-simple-0.6.2/docs/src/Database.PostgreSQL.Simple.Types.html#Query) which is just a newtyped `ByteString`. Ideally then we would be generating ByteStrings directly. Unfortunately it doesn't seem that prettyprinter supports that. We'd have to generate `Text` and then UTF8-encode the `Text`.

That is true, and its the approach I've taken now. The nice thing is that now that we have this code creating a `prettyprinter` document we can output it to other formats. A next step would actually be to add a ByteString encoder to the `prettyprinter` package IMO. If you look at the module hierarchy here: https://hackage.haskell.org/package/prettyprinter-1.3.0, there are output modules for several formats. If we had one for ByteString, we could avoid the generated-then-encode pattern.

csasarak:
I've done my best to reduce the number of changes made to the body of those modules. I have also opened an issue with prettyprinter to gauge their interest in merging a bytestring renderer, unless you think this is the wrong approach.

tomjaguarpaw:
I pushed https://github.com/tomjaguarpaw/haskell-opaleye/tree/prettyprinter-test.  It makes zero changes in the body of the printing modules.  Now that we've got to that point feel free to build on top and go crazy optimising as much as you can!

csasarak:
Thank you! I will see what I can do.

PR #438, opened 2019-08-01
Initial benchmark for query text generation

csasarak:
I started a very small benchmark suite, for the code in #436 this can help with determining if/when things are indeed faster or not. Please let me know if you have any particular additional queries you think would be good to exercise this subsystem.


tomjaguarpaw:
Sorry that I haven't had a chance to look at this yet.  You're welcome to keep it open if you like.

tomjaguarpaw:
OK, great.  Looking forward to seeing what you come up with!

PR #440, opened 2019-08-25
added readOnly

mherzl:
I often use this `readOnly` definition for SERIAL columns which I only want to auto-increment.

Want to include it in the library?


tomjaguarpaw:
Great idea!  What are the ergonomics of this like?  I guess it just means you pass `()` instead of `Nothing` whenever you insert something?

tomjaguarpaw:
Fantastic, thanks.  Unfortunately this doesn't fit in with the functional dependency in `TableColumn` otherwise we could use it there too.  (I'm not sure that functional dependency is strictly needed).

PR #441, opened 2019-10-12
Do not recommend using Opaleye.Constant

chris-martin:
The docs for Opaleye.SqlTypes recommended using Opaleye.Constant instead, but the docs for Opaleye.Constant say not to use it. So I'm guessing that leaving this note in was a mistake?


tomjaguarpaw:
Ah, thank you for reporting this.  Actually it should recommend `Opaleye.ToFields`.  I fixed that here: https://github.com/tomjaguarpaw/haskell-opaleye/commit/bb351f611b5c406be82647af755bac87b4d5ff41

chris-martin:
Ah, I see, excellent.

PR #442, opened 2019-10-15
Skip the tests that require Postgres by default

chris-martin:
I'm running into trouble building Opaleye as a Nix package because the tests fail when there is no running Postgres server, so I was thinking it might be nice if these tests were disabled by default and re-enabled by a flag.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE Arrows              #-}
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
Thanks Chris.  I'd like to minimise the additional complexity to Opaleye as much as possible.  My first question is, would it be satisfactory to simply disable tests when building Opaleye for Nix?  If not, my second proposal would be to add a command-line flag to the test suite which enables or disables the Postgres tests.  Would that be satisfactory?  I'm not sure how setting that flag would be done in the cabal file.

Do any other Nix users want to chime in?  @ocharles perhaps?

expipiplus1:
```haskell
-- From https://github.com/input-output-hk/ouroboros-network
-- | Infer from environment variables whether we are running within a Nix build
-- (and not just a nix-shell).
inNixBuild :: IO Bool
inNixBuild = do
  let testEnv = fmap (maybe False (not . null)) . System.Environment.lookupEnv
  haveNixBuildDir <- testEnv "NIX_BUILD_TOP"
  inNixShell      <- testEnv "IN_NIX_SHELL"
  pure (haveNixBuildDir && not inNixShell)
```

Something like this can be used to determine when to run such tests.

Issue #451, opened 2020-01-27
Type Family Nulls with N column


jaredramirez:
Hey, first of all thanks for the great library! I'm really liking the ability to define tables with type families, it really reduces the the possibility for bugs compared to fully polymorphic types, especially with big tables!

I've run into an issue with using `Nulls` with `N`ullalbe columns in a table.

I have a table setup like so:
```
data LocationT f
  = Location
      { ...
        sortTypes :: TableRecordField f [Text] (SqlArray SqlText) N Opt,
        ...
      }

table :: Table (LocationT W) (LocationT O)
table = ...
```

And tried to define a partially applied `leftJoinA` function like so:
```
leftJoin ::
  Opaleye.SelectArr
    ( LocationT O ->
      Opaleye.Field Opaleye.SqlBool
    )
    (LocationT Nulls)
leftJoin =
  Opaleye.leftJoinA (Opaleye.selectTable table)
```

This yields the error:
```
 • No instance for (D.Default
                         Opaleye.Internal.Join.NullMaker
                         (Opaleye.Column (Opaleye.Nullable (Opaleye.PGArray SqlText)))
                         (Opaleye.Internal.TypeFamilies.A
                            ('Opaleye.Internal.TypeFamilies.H
                               Opaleye.Internal.TypeFamilies.NullsT)
                            ('Opaleye.Internal.TypeFamilies.C '([Text], SqlArray SqlText, N))))
        arising from a use of ‘Opaleye.leftJoinA’
    • In the expression: Opaleye.leftJoinA (Opaleye.selectTable table)
      In an equation for ‘leftJoin’:
          leftJoin = Opaleye.leftJoinA (Opaleye.selectTable table)
```

I looked at the source of `Opaleye.Internal.TypeFamilies`, and saw that there was no type instance for `NullsT` with `N`, so I added an instance:
```
import qualified Opaleye.Internal.TypeFamilies as TF

type instance TF.A ('TF.H TF.NullsT) ('TF.C '(h, o, TF.N)) = Opaleye.Column (Opaleye.Nullable o)
```
and it compiles fine.

It seems like an easy fix, but I'm not sure if this is the correct way for this to be fixed or if there's a reason that this type instance doesn't exist in the first place.

tomjaguarpaw:
Oh, well spotted! I think that's just an oversight on my part.  The following line should actually be independent of its nullability argument

[...]

PR #452, opened 2020-01-28
Fix Nulls type family bug

jaredramirez:
Fixes https://github.com/tomjaguarpaw/haskell-opaleye/issues/451


tomjaguarpaw:
Thanks!  Don't worry about the test failure.  I'll have to push something to master to fix that.  It looks like a Stack file crept in there by accident though.  Could you please remove it?

jaredramirez:
Oops, didn't see that. I've removed it

tomjaguarpaw:
Thanks, I squashed and merged via the command line.

PR #458, opened 2020-04-09
Add LeftJoinLateral as a JoinType

duairc:
I don't expose this anywhere because I'm actually using Rel8, which only uses opaleye to generate SQL. If you're interested what motivated this however, it's so I can make this combinator:

```haskell
optional :: QueryArr a b -> QueryArr a (MaybeTable b)
optional query = O.QueryArr arrow
  where
    arrow (a, left, tag) = (MaybeTable t' b, join, O.next tag')
      where
        join = O.Join O.LeftJoinLateral true [] bindings left right
        (MaybeTable t b, right, tag') = f (a, O.Unit, tag)
          where
            O.QueryArr f = MaybeTable <$> pure (lit (Just False)) <*> query
        (t', bindings) = O.run (O.runUnpackspec unpackColumns (O.extractAttr "maybe" tag') t)
    true = case lit True of Expr t -> t
```

### DIFF
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -18,7 +18,7 @@ data BinOp = Except
            | IntersectAll
              deriving Show
 
-data JoinType = LeftJoin | RightJoin | FullJoin deriving Show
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
Cool, thanks!  At some point I should get around to supporting lateral joins properly, although people don't seem completely desperate to have them.

PR #459, opened 2020-04-17
Add option for 'ON CONFLICT DO UPDATE'

poscat0x04:


tomjaguarpaw:
Thanks! I never used this feature though so I'm unsure about how it is supposed to be used.  [The `INSERT` docs](https://www.postgresql.org/docs/current/sql-insert.html) suggest that `DO UPDATE` should always be followed by `SET` of some column names. Is it possible to have `DO UPDATE` on its own?

src/Opaleye/Internal/HaskellDB/Sql.hs
```diff
@@ -65,6 +65,7 @@ data SqlUpdate  = SqlUpdate SqlTable [(SqlColumn,SqlExpr)] [SqlExpr]
 data SqlDelete  = SqlDelete SqlTable [SqlExpr]
 
 data OnConflict = DoNothing
+                | DoUpdate
                 -- ^ @ON CONFLICT DO NOTHING@
```
tomjaguarpaw, line 69: The comment has become attached to the wrong constructor.

poscat0x04:
oops. That was dumb. I should've read the manual before making the pull request. Sorry for the trouble I've caused.

tomjaguarpaw:
That's no problem at all.

PR #460, opened 2020-05-10
Add InnerJoinLateral as a JoinType

duairc:
This time my motivation is to able to make the following combinator:

```haskell
laterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
laterally transform as = O.QueryArr go
  where
    go (i, left, tag) = (b, join, tag')
      where
        O.QueryArr f = transform (as . pure i)
        (b, right, tag') = f ((), O.Unit, tag)
        join = O.Join O.InnerJoinLateral true [] [] left right
    true = case lit True of Expr t -> t
```

### DIFF
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -18,7 +18,7 @@ data BinOp = Except
            | IntersectAll
              deriving Show
 
-data JoinType = LeftJoin | RightJoin | FullJoin | LeftJoinLateral deriving Show
+data JoinType = LeftJoin | RightJoin | FullJoin | InnerJoinLateral | LeftJoinLateral deriving Show
 
 data TableIdentifier = TableIdentifier
   { tiSchemaName :: Maybe String
--- a/src/Opaleye/Internal/Print.hs
+++ b/src/Opaleye/Internal/Print.hs
@@ -95,6 +95,7 @@ ppJoinType :: Sql.JoinType -> Doc
 ppJoinType Sql.LeftJoin = text "LEFT OUTER JOIN"
 ppJoinType Sql.RightJoin = text "RIGHT OUTER JOIN"
 ppJoinType Sql.FullJoin = text "FULL OUTER JOIN"
+ppJoinType Sql.InnerJoinLateral = text "INNER JOIN LATERAL"
 ppJoinType Sql.LeftJoinLateral = text "LEFT OUTER JOIN LATERAL"
 
 ppAttrs :: Sql.SelectAttrs -> Doc
--- a/src/Opaleye/Internal/Sql.hs
+++ b/src/Opaleye/Internal/Sql.hs
@@ -65,7 +65,7 @@ data Binary = Binary {
   bSelect2 :: Select
 } deriving Show
 
-data JoinType = LeftJoin | RightJoin | FullJoin | LeftJoinLateral deriving Show
+data JoinType = LeftJoin | RightJoin | FullJoin | InnerJoinLateral | LeftJoinLateral deriving Show
 data BinOp = Except | ExceptAll | Union | UnionAll | Intersect | IntersectAll deriving Show
 
 data Label = Label {
@@ -219,6 +219,7 @@ joinType :: PQ.JoinType -> JoinType
 joinType PQ.LeftJoin = LeftJoin
 joinType PQ.RightJoin = RightJoin
 joinType PQ.FullJoin = FullJoin
+joinType PQ.InnerJoinLateral = InnerJoinLateral
 joinType PQ.LeftJoinLateral = LeftJoinLateral
 
 binOp :: PQ.BinOp -> BinOp


tomjaguarpaw:
Sure, though we may have to revisit this later because it looks like that, grammatically speaking, `LATERAL` applies to the subquery, not the join type.

https://www.postgresql.org/docs/9.4/queries-table-expressions.html

tomjaguarpaw:
By the way, I suspect that this actually allows

```haskell
laterally :: (a -> QueryArr i b) -> QueryArr i a -> QueryArr i b
```

Lateral join induce monad structure.

tomjaguarpaw:
Thinking more, I believe that `(Query a -> Query b) -> QueryArr i a -> QueryArr i b` is impossible to write. The `QueryArr i a` can depend on its `i`and I'm not sure therefore how it can be transformed by something of type `Query a -> Query b`.  On the other hand `QueryArr a b -> QueryArr i a -> QueryArr i b` is possible, but then one may as well go the whole hog and use `(a -> QueryArr i b) -> QueryArr i a -> QueryArr i b`. I have type checked the following, but not tested it.

```haskell
{-# OPTIONS_GHC -Wall #-}

module Lateral where

import Opaleye
import Opaleye.Internal.QueryArr as O
import Opaleye.Internal.PrimQuery as O
import Opaleye.Internal.HaskellDB.PrimQuery as O

laterally :: (a -> QueryArr i b) -> QueryArr i a -> QueryArr i b
laterally transform as = simpleQueryArr go
  where
    go (i, startTag) = (columnsB, primQueryJoin, endTag)
      where
        (columnsA, primQueryA, midTag) =
          runSimpleQueryArr as (i, startTag)
        (columnsB, primQueryB, endTag) =
          runSimpleQueryArr (transform columnsA) (i, midTag)

        true = O.ConstExpr (BoolLit True)
        primQueryJoin = O.Join innerJoinLateral true [] [] primQueryA primQueryB

        innerJoinLateral = error "Implement InnerJoinLateral"
```

Thanks for initiating work on lateral joins. It's been missing for a long time!

shane-circuithub:
Well, thanks for merging that anyway! Are you sure that `(Query a -> Query b) -> QueryArr i a -> QueryArr i b` is impossible to write? The version I gave above not only typechecks, but I've used it to write queries that do aggregations, distinct on, orders, limits, etc, and the SQL it produces is valid and does exactly what I want it to (and fast).

shane-circuithub:
I'm aware that `LATERAL` also gives you a `>>=`-type operation — I wasn't proposing `laterally` in place of that, I mentioned it just because that's the one I need right now. It basically takes away the restrictions around using aggregations and similar on arbitrary `QueryArr`s.

shane-circuithub:
And yes, I think you're right about the grammar, I realised that after I'd made the first commit. I tried to do it properly but the first few attempts I made didn't work. I'm not an expect on Opaleye internals, I mostly just poke it until the types fit, and then poke it again until the SQL is valid :)

shane-circuithub:
I don't think the function you have is sound by the way — I think `bind` would have to be restricted just to `Query`. For example. you can implement a function with the type of my `laterally` in terms of such a `bind` like:

```haskell
laterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
laterally transform = bind (dimap pure (dimap mempty id) transform)
```

But it produces invalid SQL when used with aggregations.

tomjaguarpaw:
The grammar issue doesn't affect the public API so it's fine to leave it for now.

> The version I gave above not only typechecks

Ah, I see, `.` is `Control.Category..`.  That makes more sense than `Prelude..`. It also seems to use `Expr` and `lit` from rel8, but that's not an essential difficulty.  With those pieces, yes, I can see now how you can write it!

duairc:
Ah, yes, sorry. Obviously that could easily be `O.ConstExpr (BoolLit True)` — I'm working with Rel8 so I was just copy-pasting what I have. I've been experimenting a lot with stuff like this in the last few months. I'm almost (but not quite yet) ready to declare those experiments a success. But long story short, when I get a chance to clean up all of this, expect some more patches :)

tomjaguarpaw:
Cool, very exciting!

tomjaguarpaw:
Aha, I think I understand what I got wrong.  It was the `i` that you wanted to reuse from the left query, not the `a`!  In which case I suspect this is a more useful combinator and might actually work!

```haskell
laterally1 :: (i -> Query b) -> QueryArr i b
laterally1 transform = O.QueryArr go
  where
    go (i, primQueryL, tag) = (b, primQueryJoin, tag')
      where
        (b, primQueryR, tag') = runSimpleQueryArr (transform i) ((), tag)
        primQueryJoin = O.Join innerJoinLateral true [] [] primQueryL primQueryR

    true = O.ConstExpr (BoolLit True)
    innerJoinLateral = error "Implement InnerJoinLateral"
```

duairc:
Haha, yes, that was actually the original combinator I came up with, except I called it `lateral`, and you can write `laterally` in terms of it quite easily. However, you can also write a kind of `bind` in terms of it:

```haskell
bind :: QueryArr i a -> (a -> Query b) -> QueryArr i b
bind as f = lateral f . aslaterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
```

And then can write `laterally` in terms of that:

```haskell
laterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
laterally f = flip bind (f . pure)
````

However, although it has the same type, it's not the same function, and if you use it for aggregations, it won't work, you'll get errors like `ERROR:  aggregate functions are not allowed in FROM clause of their own query level`. I guess the main point I'm making is that that `lateral`, at least as you've written it, allows bad things.

That isn't to say that you can't write a correct `laterally` in terms of `lateral`, you can:

```haskell
laterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
laterally f as = lateral (f . (`lmap` as) . const)
```

And the implementation above is just that with the definition of `lateral` inlined.

tomjaguarpaw:
Aha, could you give me an example of an Opaleye expression using `laterally1` that generates crashing SQL?

tomjaguarpaw:
This is valid, at least according to [DB Fiddle](https://www.db-fiddle.com/f/4dmkvUunSQk3A2S23K3J4i/0).  Perhaps `laterally1` breaks on some more complicated examples.  If you can construct a query using `laterally1` which breaks then please let me know.

```haskell
table1 :: Table (Field SqlInt4, Field SqlInt4) (Field SqlInt4, Field SqlInt4)
table1 = table "table1" (p2 (tableField "table1column1"
                            ,tableField "table1column2"))

example :: Query (Field SqlInt4)
example = proc () -> do
  (f1, _) <- selectTable table1 -< ()
  f <- laterally1 (\f1' -> Opaleye.aggregate Opaleye.sum $ proc () -> do
           (f1'',f2'') <- selectTable table1 -< ()
           Opaleye.restrict -< f1'' .== f1'
           returnA -< f2''
       ) -< f1
  returnA -< f

showExample :: IO ()
showExample = mapM_ putStrLn $ showSql example
```

```sql
SELECT
"result0_3" as "result1_4"
FROM (SELECT *
      FROM
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   "table1column1" as "table1column10_1",
                   "table1column2" as "table1column21_1"
                   FROM "table1" as "T1") as "T1") as "T1") as "T1"
      INNER JOIN LATERAL
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   SUM("table1column21_2") as "result0_3"
                   FROM (SELECT
                         *
                         FROM (SELECT
                               "table1column1" as "table1column10_2",
                               "table1column2" as "table1column21_2"
                               FROM "table1" as "T1") as "T1"
                         WHERE (("table1column10_2") = ("table1column10_1"))) as "T1"
                   GROUP BY COALESCE(0)) as "T1") as "T1") as "T2"
      ON
      TRUE) as "T1"
```

duairc:
Okay, here's the Haskell code I have

```haskell
lateral :: (i -> Query b) -> QueryArr i b
lateral transform = O.QueryArr go
  where
    go (i, primQueryL, tag) = (b, primQueryJoin, tag')
      where
        (b, primQueryR, tag') = O.runSimpleQueryArr (transform i) ((), tag)
        primQueryJoin = O.Join O.InnerJoinLateral true [] [] primQueryL primQueryR
    true = O.ConstExpr (O.BoolLit True)

bind :: QueryArr i a -> (a -> Query b) -> QueryArr i b
bind as f = lateral f . as

laterally :: (Query a -> Query b) -> QueryArr i a -> QueryArr i b
laterally f = flip bind (f . pure)

table1 :: Query (O.Field O.SqlInt4, O.Field O.SqlInt4)
table1 = O.values
  [ (1, 1)
  , (1, 2)
  , (3, 3)
  , (3, 4)
  , (5, 5)
  , (5, 6)
  ]

example :: Query (O.Field O.SqlInt4)
example = snd <$> laterally (O.aggregate (p2 (O.groupBy, O.sum))) table1
```

That generates the following invalid SQL:

```sql
SELECT
"result1_2" as "result1_3"
FROM (SELECT *
      FROM
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT "column1" as "values0_1",
                          "column2" as "values1_1"
                   FROM
                   (VALUES
                    (CAST(1 AS integer),CAST(1 AS integer)),
                    (CAST(1 AS integer),CAST(2 AS integer)),
                    (CAST(3 AS integer),CAST(3 AS integer)),
                    (CAST(3 AS integer),CAST(4 AS integer)),
                    (CAST(5 AS integer),CAST(5 AS integer)),
                    (CAST(5 AS integer),CAST(6 AS integer))) as "V") as "T1") as "T1") as "T1"
      INNER JOIN LATERAL
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   "values0_1" as "result0_2",
                   SUM("values1_1") as "result1_2"
                   FROM (SELECT
                         0) as "T1"
                   GROUP BY "values0_1") as "T1") as "T1") as "T2"
      ON
      CAST(TRUE AS boolean)) as "T1"
```

It would seem that the definition `laterally` above is incorrect and that the correct definition should be `laterally f as = lateral (f . (as .) . pure)`. However, the point is that I was able to generate invalid SQL using `lateral`.

duairc:
Actually, the above example is less than ideal, because `table1` is just a `Query`, but the whole point of my `laterally` is that it will work on `QueryArr`s too. So here's a modified version that does that:

```haskell
table1 :: Query (O.Field O.SqlInt4, O.Field O.SqlInt4)
table1 = O.values
  [ (1, 1)
  , (1, 2)
  , (3, 3)
  , (3, 4)
  , (5, 5)
  , (5, 6)
  ]

tableA :: QueryArr (O.Field O.SqlInt4) (O.Field O.SqlInt4)
tableA = proc i -> do
  (a, b) <- table1 -< ()
  O.restrict -< a O..== i
  returnA -< b

exampleA :: QueryArr (O.Field O.SqlInt4) (O.Field O.SqlInt4)
exampleA = laterally (O.aggregate O.sum) tableA

example :: Query (O.Field O.SqlInt4)
example = exampleA . O.values [1, 3, 5]
```

With the good `laterally`, it generates the following (valid) SQL:

```sql
SELECT
"result0_3" as "result1_4"
FROM (SELECT *
      FROM
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT "column1" as "values0_1"
                   FROM
                   (VALUES
                    (CAST(1 AS integer)),
                    (CAST(3 AS integer)),
                    (CAST(5 AS integer))) as "V") as "T1") as "T1") as "T1"
      INNER JOIN LATERAL
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   SUM("values1_2") as "result0_3"
                   FROM (SELECT
                         *
                         FROM (SELECT "column1" as "values0_2",
                                      "column2" as "values1_2"
                               FROM
                               (VALUES
                                (CAST(1 AS integer),CAST(1 AS integer)),
                                (CAST(1 AS integer),CAST(2 AS integer)),
                                (CAST(3 AS integer),CAST(3 AS integer)),
                                (CAST(3 AS integer),CAST(4 AS integer)),
                                (CAST(5 AS integer),CAST(5 AS integer)),
                                (CAST(5 AS integer),CAST(6 AS integer))) as "V") as "T1"
                         WHERE (("values0_2") = ("values0_1"))) as "T1"
                   GROUP BY COALESCE(0)) as "T1") as "T1") as "T2"
      ON
      CAST(TRUE AS boolean)) as "T1"
```

With the bad `laterally` (in terms of `bind`), it generates this (invalid) SQL:

```sql
SELECT
"result0_3" as "result1_4"
FROM (SELECT *
      FROM
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT "column1" as "values0_1"
                   FROM
                   (VALUES
                    (CAST(1 AS integer)),
                    (CAST(3 AS integer)),
                    (CAST(5 AS integer))) as "V") as "T1",
                  (SELECT "column1" as "values0_2",
                          "column2" as "values1_2"
                   FROM
                   (VALUES
                    (CAST(1 AS integer),CAST(1 AS integer)),
                    (CAST(1 AS integer),CAST(2 AS integer)),
                    (CAST(3 AS integer),CAST(3 AS integer)),
                    (CAST(3 AS integer),CAST(4 AS integer)),
                    (CAST(5 AS integer),CAST(5 AS integer)),
                    (CAST(5 AS integer),CAST(6 AS integer))) as "V") as "T2"
             WHERE (("values0_2") = ("values0_1"))) as "T1") as "T1"
      INNER JOIN LATERAL
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   SUM("values1_2") as "result0_3"
                   FROM (SELECT
                         0) as "T1"
                   GROUP BY COALESCE(0)) as "T1") as "T1") as "T2"
      ON
      CAST(TRUE AS boolean)) as "T1"
```

tomjaguarpaw:
Perfect, thanks.  Here's a somewhat simplified version and its (crashing) SQL.  I suppose that the hypothesis that LATERAL allows one to use a monad interface was incorrect all along.  Aggregation does still seem to get in the way.

```haskell
example :: Query _
example = laterally1 (Opaleye.aggregate Opaleye.sum . pure)
          <<< Opaleye.values [0 :: Field SqlInt4]
```

```sql
SELECT
"result0_2" as "result1_3"
FROM (SELECT *
      FROM
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT "column1" as "values0_1"
                   FROM
                   (VALUES
                    (CAST(0 AS integer))) as "V") as "T1") as "T1") as "T1"
      INNER JOIN LATERAL
      (SELECT
       0,
       *
       FROM (SELECT
             *
             FROM (SELECT
                   SUM("values0_1") as "result0_2"
                   FROM (SELECT
                         0) as "T1"
                   GROUP BY COALESCE(0)) as "T1") as "T1") as "T2"
      ON
      TRUE) as "T1"
```

tomjaguarpaw:
Hmm, perhaps it may be possible with a bit of thought.  The former is forbidden but the latter is allowed.  Naturally they ought to be the same!  I wonder if we might be able to use this to find a workaround.

```sql
                   SUM("values0_1") as "result0_2"
                   FROM (SELECT
                         0) as "T1"
```

```sql
                   SUM("foo") as "result0_2"
                   FROM (SELECT
                         "values0_1" as "foo",
                         0) as "T1"
```

duairc:
For what it's worth here's another operator you can get (in terms of my `laterally`) that's very nice for lifting things like `UNION`s into arbitrary `QueryArr`s:

```haskell
bilaterally :: (Query a -> Query b -> Query c) -> QueryArr i a -> QueryArr i b -> QueryArr i c
bilaterally transform as bs =
  laterally (liftA2 transform (fmap fst) (fmap snd)) (liftA2 (,) as bs)
```

tomjaguarpaw:
An [`Aggregator` already contains stuff that's at least as powerful as an `Unpackspec`](https://hackage.haskell.org/package/opaleye-0.6.7004.1/docs/src/Opaleye.Internal.Aggregate.html#Aggregator), I believe.  We can either add the ability to add extra columns to the aggregated table to the `Aggregator` or piggy back on something it already contains, to do the same job.  I'd say I'm 95% certain this way will work (unless there's some other Postgres syntactic restriction I'm not aware of).

tomjaguarpaw:
`bilaterally` seems a little suspicious.  It takes the product of the two `QueryArr`s!  Can you show it doing a `unionAll` on a simple example?

duairc:
D'oh, you're right about `bilaterally`, that product is dubious. To be honest the definition I've been using is actually the following (which does work):

```
bilaterally :: (Query a -> Query b -> Query c) -> QueryArr i a -> QueryArr i b -> QueryArr i c
bilaterally transform as bs = O.QueryArr go
  where
    go (i, left, tag) = (c, join, tag')
      where
        O.QueryArr f = transform (as . pure i) (bs . pure i)
        (c, right, tag') = f ((), O.Unit, tag)
        join = O.Join O.InnerJoinLateral true [] [] left right
    true = case lit True of Expr t -> t
```

I was just trying to be clever and define it in terms of `laterally` but at least that first attempt wouldn't work.

duairc:
But that's really exciting that you think `aggregate` could be massaged into something that could work. I would really love if the type of `lateral` could be `(a -> Query b) -> QueryArr a b`, that was the original type I came up with.

duairc:
Okay, `bilaterally` can be easily written in terms of `lateral`:

```haskell
bilaterally :: (Query a -> Query b -> Query c) -> QueryArr i a -> QueryArr i b -> QueryArr i c
bilaterally transform as bs = lateral (\i -> transform (as . pure i) (bs . pure i))
```

So if we can sort out the aggregations then it really seems that that's the fundamental operator that we want!

tomjaguarpaw:
> bilaterally can be easily written in terms of lateral

That's very good news. It makes a working version of `lateral` into an exciting target.

> But that's really exciting that you think aggregate could be massaged into something that could work. I would really love if the type of lateral could be (a -> Query b) -> QueryArr a b, that was the original type I came up with.

I'm optimistic for a few reasons.  Firstly, we know what the semantics of `lateral` should be. It basically makes `Query`into a monad, which is super useful.  This monad instance has been hypothesised ever since Postgres started supporting `LATERAL`. Secondly, I can see exactly why `lateral` didn't work in the case of the aggregation example and how to fix it. It's a weakness in Postgresql, unfortunately, but I've worked around many such weaknesses in the past. We also need to be sure that all other primitives of type `Query a1 -> ... Query an -> Query r` behave properly in conjunction with `LATERAL`, but I'm fairly confident that they will. The list includes `DISTINCT` (which is essentially an aggregation) and all binary operations.

ocharles:
I don't have much to add, just that this is all very exciting. I thought `LATERAL` gave us a monad ages ago, but then came across the counter example of `SELECT * FROM (SELECT a FROM t) INNER JOIN LATERAL (SELECT (sum(t.a))` and thought that was probably the end of the road. But it appears not! I look forward to seeing what comes next

duairc:
Just for reference, here's `>>=` in terms of `lateral`:

```haskell
bind :: QueryArr i a -> (a -> QueryArr i b) -> QueryArr i b
bind as f = lateral (\(i, a) -> f a . pure i) . liftA2 (,) id as
```

and `app` (from `ArrowApply`):

```haskell
app :: QueryArr (QueryArr i a, i) a
app = lateral (\(f, i) -> f . pure i)
```

And the reverse (`lateral` in terms of `>>=`)

```haskell
lateral :: (i -> Query a) -> QueryArr i a
lateral'' f = bind id (\i -> f i . pure ())
```

and `lateral` in terms of `app`:

```haskell
lateral :: (i -> Query a) -> QueryArr i a
lateral' f = app . arr (\i -> (f i . pure (), i))
```

So these are definitely all the same thing.

tomjaguarpaw:
I [asked about this limitation on dba.stackexchange](https://dba.stackexchange.com/questions/266988/why-is-it-that-aggregate-functions-are-not-allowed-in-from-clause-of-their-own).

ocharles:
There's a chance you might get a more detailed response on the PG mailing list too, if this doesn't work out.

tomjaguarpaw:
Posted to postgresql-general: https://www.postgresql.org/message-id/20200513110251.GC24083%40cloudinit-builder

I'm mostly interested in confirming whether my planned transformation really is one that gives us correct behaviour, although we can probably deduce that from our own experiments.

tomjaguarpaw:
On postgresql-general the seem to be disinclined to comment on a semantics that doesn't actually exist, which seems understandable.  Anyway, I'm fairly sure it will work and I hope to write up the design for doing this fairly soon.

duairc:
Yeah, I've been following the discussions. I just wanted to say that at least empirically, I'm happy that your transformation gives the results I would expect it to in all the cases that I've tried it.

PR #463, opened 2020-05-12
Implement LATERAL subqueries

### LINKED ISSUE #359: LATERAL JOIN support?
I was wondering if there were any plans to support lateral joins.

 I've got some complex data aggregation queries that have proven to be very difficult to express as I'd like without access to lateral join functionality.

duairc:
Implement LATERAL subqueries
    
For some context, see the dicussion on #460.
    
I define `lateral` in `Opaleye.Internal.QueryArr` so that we can use it to define `ArrowApply` and `Monad` instances for `QueryArr`.
    
In terms of the grammar, I don't introduce any specific AST constructor for `LATERAL`. Instead, when generating the SQL, we simply always write `FROM LATERAL` instead of `FROM` for subqueries. The advantage of this is that we get a lawful monad instance (`>=> return` is exactly `id` and doesn't pollute the AST). From my testing, this doesn't seem to change the plans that are by Postgres generated for non-lateral queries.
    
I re-export it from `Query.Lateral`, along with `laterally` and `bilaterally`, which also came up in the aforementioned discussion.
    
This would fix #359.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -56,6 +56,7 @@ library
                    Opaleye.FunctionalJoin,
                    Opaleye.Join,
                    Opaleye.Label,
+                   Opaleye.Lateral,
                    Opaleye.Manipulation,
                    Opaleye.Map,
                    Opaleye.Operators,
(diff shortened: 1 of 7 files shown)


duairc:
You'll probably at least want to fix [this issue with aggregates](https://github.com/tomjaguarpaw/haskell-opaleye/pull/460#issuecomment-626716160) first before merging this :)

duairc:
This was implemented in #470, so closing.

PR #465, opened 2020-05-16
Lateral

tomjaguarpaw:
@duairc I would appreciate it if you would try this out and see if you think it behaves as expected.  It passes the CI tests (but they don't test lateral, yet) and it passes the tiny example I tried.  I'd appreciate it if you would give it a more thorough testing.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -56,6 +56,7 @@ library
                    Opaleye.FunctionalJoin,
                    Opaleye.Join,
                    Opaleye.Label,
+                   Opaleye.Lateral,
                    Opaleye.Manipulation,
                    Opaleye.Map,
                    Opaleye.Operators,
--- a/src/Opaleye/Internal/Aggregate.hs
+++ b/src/Opaleye/Internal/Aggregate.hs
@@ -88,18 +88,30 @@ runAggregator (Aggregator a) = PM.traversePM a
 aggregateU :: Aggregator a b
            -> (a, PQ.PrimQuery, T.Tag) -> (b, PQ.PrimQuery, T.Tag)
 aggregateU agg (c0, primQ, t0) = (c1, primQ', T.next t0)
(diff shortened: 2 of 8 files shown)


duairc:
Hey, this looks good to me! I compiled a large codebase that uses pretty much every feature of Rel8/Opaleye against this branch, and I ran the query benchmarking suite it has, and there's no difference before and after. Which at the very least means that the SQL is valid and it hasn't impacted performance.

I also tried to some different pathological combinations of aggregations and `LATERAL`, but I can't seem to get Opaleye to generate invalid SQL anymore.

duairc:
(That codebase was already using my LATERAL patch, so the only difference would be your aggregation fix I think.)

tomjaguarpaw:
That's great!  And am I correct in thinking that your benchmarks show there is no difference in performance between `LATERAL`-using and non-`LATERAL`-using code (for queries that don't make use of the additional functionality afforded by `LATERAL`?)

duairc:
Yes, that's correct! I was very pleased when I learnt that. You can verify too with `EXPLAIN` that Postgres is generating the same plans with or without `LATERAL`.

tomjaguarpaw:
Yes, that is very pleasing indeed.

tomjaguarpaw:
I'm not sure when I'll be able to get round to finishing this off properly.  The code in this PR probably needs to be polished a bit before merge.  @duairc I'd encourage you to merge this into your branch which provides the lateral combinators, try it out, and report back your experience!

tomjaguarpaw:
>  I'd encourage you to merge this into your branch

Oh, in fact it already contains your branch, so I think in the interim if people want to try this branch out and provide experience reports that will be very helpful.

tomjaguarpaw:
Closing in favour of https://github.com/tomjaguarpaw/haskell-opaleye/pull/466

PR #466, opened 2020-05-31
Lateral

tomjaguarpaw:
Implements a Monad instance for queries, using `LATERAL` joins

(this comes with lateral, which is equivalent to the Monad instance, and laterally and bilaterally which are useful combinators)

I'm more likely to merge this than https://github.com/tomjaguarpaw/haskell-opaleye/pull/465 because the latter makes all joins lateral.  I don't see the point of that at the moment.

Once people start writing monadic queries there's no going back, so I want to be 100% sure this is a sensible thing to do.  I would welcome people testing this.

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -1,10 +1,12 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE Arrows #-}
 
 module QuickCheck where
 
 import qualified Opaleye as O
+import qualified Opaleye.Lateral as OL
 import qualified Database.PostgreSQL.Simple as PGS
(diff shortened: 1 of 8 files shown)


duairc:
The reason for making all joins lateral was so that the monad instance would be lawful, i.e., `m >>= pure = m`. It also plays much nicer with the optimisations in `Opaleye.Internal.Optimize`, i.e., the flattening of `Product`s, because `Join InnerJoinLateral` is represented differently to `Product`.

duairc:
On another note though, I did manage to generate a query the other day (using `bilaterally` if I recall correctly) that would cause Postgres to segfault when I ran it! However when I tried to shrink it down into a minimal test case it stopped doing that. In any case it sounds more like a bug in Postgres than Opaleye.

tomjaguarpaw:
Closing in favour of https://github.com/tomjaguarpaw/haskell-opaleye/pull/470

PR #468, opened 2020-06-13
Remove Bindings from Binary and introduce new PrimQuery constructor Rebind

duairc:
Basically, Opaleye implements binary operations by rewriting all the columns of both queries to have the same name, before combining. Previously this rewriting of column names was represented in the AST as part of the `Binary` operation itself, whereas now it's a separate step (`Rebind`), and `sameTypeBinOpHelper` (which `unionAll`, `exceptAll` etc are defined in terms of) will do both of these steps at once.

The reason this is necessary is because the `removeEmpty` optimisation, which removes can remove binary operations that are unnecessary (.e.g, `unionAll`ing with `values []`), would generate incorrect queries before because the column names introduced by the binary operation would no longer exist, but the expressions would still refer to them. By making the rebinding and the binary operation itself separate steps, we can safely remove the latter while keeping the former.

### DIFF
--- a/src/Opaleye/Internal/Binary.hs
+++ b/src/Opaleye/Internal/Binary.hs
@@ -47,7 +47,10 @@ sameTypeBinOpHelper binop binaryspec q1 q2 = Q.simpleQueryArr q where
             PM.run (runBinaryspec binaryspec (extractBinaryFields endTag)
                                     (columns1, columns2))
 
-          newPrimQuery = PQ.Binary binop pes (primQuery1, primQuery2)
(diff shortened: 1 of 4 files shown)


tomjaguarpaw:
This is great!  I have added it to my `binary` branch.

duairc:
I'm closing this because it appears to be in master already.

tomjaguarpaw:
Thanks. Yes, I did merge it into `master`. Normally GitHub notices that and closes the PR.  Seems like it didn't notice in this case.  `Rebind` is a very useful addition.

PR #469, opened 2020-06-17
Implement EXISTS directly

duairc:
We already have restrictExists and restrictNotExists, which implement `WHERE EXISTS` and `WHERE NOT EXISTS` respectively, however we can't directly use `EXISTS` to get a `Column PGBool`.

My first attempt at this added `exists` and then redefined `restrictExists` and `restrictNotExists` in terms of `laterally exists`. That basically meant that code that originally looked like this:

```sql
SELECT * FROM
  (SELECT "result0_1"
   FROM
   ...) "T1"
WHERE EXISTS (SELECT "result0_1") as "T2"
```

would now become:

```sql
SELECT * FROM
  (SELECT "result0_1"
   FROM
   ...) "T1",
  LATERAL
  (SELECT EXISTS (SELECT foo
                 ) as "exists0_2") "T1"
WHERE "exists0_2"
```

Unfortunately, in a lot of cases, this made Postgres generate a much worse plan than before. So I've preserved the old constructor in `PrimQuery` (but renamed it to `Semijoin`), and repurposed `Exists` to implement just `EXISTS`.

I also reimplmented `inSelect` in terms of this new `exists` operator, to give a demonstration of its utility.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -52,6 +52,7 @@ library
                    Opaleye.Column,
                    Opaleye.Constant,
                    Opaleye.Distinct,
+                   Opaleye.Exists,
                    Opaleye.Field,
                    Opaleye.FunctionalJoin,
                    Opaleye.Join,
--- a/src/Opaleye/Exists.hs
+++ b/src/Opaleye/Exists.hs
@@ -1,0 +1,21 @@
+module Opaleye.Exists (exists) where
+
+import           Opaleye.Field (Field)
+import           Opaleye.Internal.Column (Column (Column))
+import           Opaleye.Internal.QueryArr (runSimpleQueryArr, simpleQueryArr)
+import           Opaleye.Internal.PackMap (run, extractAttr)
+import           Opaleye.Internal.PrimQuery (PrimQuery' (Exists))
+import           Opaleye.Internal.Tag (next)
+import           Opaleye.Select (Select)
+import           Opaleye.SqlTypes (SqlBool)
+
+-- | True if any rows are returned by the given query, false otherwise.
+--
+-- This operation is equivalent to Postgres's @EXISTS@ operator.
+exists :: Select a -> Select (Field SqlBool)
+exists q = simpleQueryArr (f . runSimpleQueryArr q)
+  where
+    f (_, query, tag) = (Column result, Exists binding query, tag')
+      where
+        (result, [(binding, ())]) = run (extractAttr "exists" tag ())
+        tag' = next tag
--- a/src/Opaleye/Internal/Optimize.hs
+++ b/src/Opaleye/Internal/Optimize.hs
@@ -10,7 +10,7 @@ import           Opaleye.Internal.Helpers   ((.:))
 import qualified Data.List.NonEmpty as NEL
 import           Data.Semigroup ((<>))
 
-import           Control.Applicative ((<$>), (<*>), pure)
+import           Control.Applicative ((<$>), (<*>), liftA2, pure)
 import           Control.Arrow (first)
 
 optimize :: PQ.PrimQuery' a -> PQ.PrimQuery' a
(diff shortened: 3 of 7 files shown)


duairc:
(This is based off of my lateral branch, which is in PR #463, but still needs your aggregate fixes.)

i'm not delighted with this and i can understand if you wouldn't be keen on merging it. I had really hoped to be able to write `restrictExists` and `restrictNotExists` in terms of `laterally exists`. If you want to see what that looks like, then check out my [`exists2` branch](https://github.com/duairc/opaleye/blob/exists2/src/Opaleye/Operators.hs#L55-L61). However, some of our queries are up to 10 times slower with that change, so we need the existing formulation as well, which negates most of the benefit of it.

I still think there's some value in this, as you can see from the new implementation of `inSelect`. But you can simulate `exists` with the combination of `optional` and `restrictExists` already, so this isn't essential (although it is slightly faster).

I'll leave this here anyway for other people to study if nothing else.

tomjaguarpaw:
I had a go at rebasing this on my `exists` branch.  If I got the conflict resolutions correct then perhaps you could force push it to your `exists` branch too.

duairc:
I've taken your `exists` branch and rebased it onto `master` and force-pushed that to my `exists` branch.

tomjaguarpaw:
Thanks, this is nicely written.  As I understand it, this does a couple of things

1. Exposes `exists :: Select a -> Select (Field SqlBool)`
2. Implements `Exists` and `SemiJoin` (WHERE /NOT EXISTS) more nicely in the backend

Have you tried writing `exists` in terms of `inSelect`?  I'd be interested to know if there is a downside to that (perhaps there's a performance downside).

duairc:
I've done something which is similar in spirit. This is how we define `exists`:

```haskell
exists :: QueryArr a b -> QueryArr a (Expr Bool)
exists = fmap (not_ . isTableNull) . optional . with
```

Where `isTableNull` and `optional` are as we discussed at ZuriHac, and `with` is something like `id *> restrictExists`.

This performs acceptably but not quite as good as having `EXISTS` natively. But my main motivation for writing this wasn't performance, but because I hoped to unify `EXISTS` with `WHERE EXISTS`/`WHERE NOT EXISTS`. But that turned out not to be practical, so I actually don't have a strong motive for this anymore, but I figured I'd throw it up anyway.

tomjaguarpaw:
OK, well it's good to have it to look at, at least, and it implements existing things in a less round about way, so might be worth merging just for that.  All my Opaleye brainpower is currently dedicated to merging `optional`/`MaybeFields`s related stuff though ...

ocharles:
Hey @tomjaguarpaw - any update on this? It would be great if we could get back to using `master`! We've been running this in production for most of 2020 with no problems.

tomjaguarpaw:
Happy to merge this. @duairc said he didn't have a strong motive for it anymore but if you do use it and it is successful in practice then let's get it in! I've rebased, fixed (minor) conflicts and added a test. I pushed it to `tomjaguarpaw/exists` and opened a PR: https://github.com/tomjaguarpaw/haskell-opaleye/pull/494

Either @duairc can reset his branch to that one or we can just merge this on this new PR.

Either way, can @ocharles and @duairc confirm that what we're about to merge agrees with what you've been using in production?

shane-circuithub:
Hey, sorry, don't merge this :) This is not the version we've been using in production. Let me update this PR quickly.

duairc:
Hey, sorry, don't merge this :) This is not the version we've been using in production. Let me just update this first :)

And yes, at the time I didn't have a strong motive for it, but I recently wrote a bunch of performance sensitive queries where having this can actually shave a second or two off the total runtime, so I would definitely like to have it.

tomjaguarpaw:
> Let me just update this first :)

OK great. Please rebase what you have been using onto master and add the property test from my branch (https://github.com/tomjaguarpaw/haskell-opaleye/pull/494/files#diff-c560e91c0e33713ca88b74fe5c4574b1ac0861a90d9f32359bd115ddd1c9f7a7)

Let's keep using this PR rather than the one I just opened.

duairc:
Actually, I think this is actually okay to merge. The version we're using in production depends on #480, and this PR is already what happens when you disentangle it from that. I'll rebase it onto your master branch and then it should be good to go.

tomjaguarpaw:
Oh, I did the rebasing so we can just reopen https://github.com/tomjaguarpaw/haskell-opaleye/pull/494

duairc:
Okay, well sorry for the confusion in that case :)

tomjaguarpaw:
No problem!

tomjaguarpaw:
Merged by https://github.com/tomjaguarpaw/haskell-opaleye/pull/494

PR #470, opened 2020-06-17
Lateral

tomjaguarpaw:
Implement lateral subqueries, but don't expose them in the public API yet.

@duairc I'd appreciate it if you would take a look and check it looks sensible to you. I'm minded to merge it ASAP.

Where you had `PrimQuery` and `[(Lateral, PrimQuery)]` I've just used `NonEmpty (Lateral, PrimQuery)`.  The first `Lateral` is indeed redundant but it makes the diff much more understandable if it is present.  I also tidied up some of the SQL generation.

(I'm going to be much happier about merging this if we make it clear that `lateral` is for intrepid adventurers only).

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -2,10 +2,12 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE Arrows #-}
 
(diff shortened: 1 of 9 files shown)


tomjaguarpaw:
(@duairc's `Rebind` was really useful for this, by the way)

duairc:
I've looked over this and it does indeed look sensible to me. I agree that it'd be good to get this merged as soon as possible!

PR #473, opened 2020-06-27
Lateral

tomjaguarpaw:


tomjaguarpaw:
@duairc @ocharles It would be good to know if [Opaleye.Internal.MaybeFields](https://github.com/tomjaguarpaw/haskell-opaleye/pull/473/files#diff-f6adf9be1d69f56be9a0864386c8e527) gives you everything you need to remove LATERAL-specific code from Rel8 and just use Opaleye directly.

[The PR is rather hefty but I don't think any of the rest of it is directly relevant]

PR #480, opened 2020-08-31
Make all queries LATERAL all the time

duairc:
[There](https://github.com/tomjaguarpaw/haskell-opaleye/pull/463) [have](https://github.com/tomjaguarpaw/haskell-opaleye/pull/465) [been](https://github.com/tomjaguarpaw/haskell-opaleye/pull/466) [multiple](https://github.com/tomjaguarpaw/haskell-opaleye/pull/470) [attempts](https://github.com/tomjaguarpaw/haskell-opaleye/pull/473) at implementing `LATERAL` in Opaleye. The one that eventually got merged, [this one](https://github.com/tomjaguarpaw/haskell-opaleye/pull/470) is different to the [original implementation](https://github.com/tomjaguarpaw/haskell-opaleye/pull/463) in that the original made all joins `LATERAL` indiscriminately (and thus the actual `lateral` combinator was a no-op that just opened up the `QueryArr` type), whereas the new one will only add `LATERAL` to queries when you specifically ask it to.

In the original implementation, [`optional`](https://hackage.haskell.org/package/opaleye-0.7.0.0/docs/Opaleye-MaybeFields.html#v:optional), the underlying `LEFT JOIN` is implemented as `LEFT JOIN LATERAL`, wheras the in new implementation, the left join is implemented as a non-lateral `LEFT JOIN`, with the laterality being added one level up. On the vast majority of queries this performs exactly the same, but there are some pathological cases where this is a regression, unfortunately. In our codebase we have observed some queries that are two orders of magnitude slower now, and several more by one order of magnitude.

I also apologise for not speaking up about this before you released Opaleye 0.7.0.0 — we were vaguely aware that this was causing us problems, but we were focused elsewhere at the time and we also didn't understand the full extent of it until we we ported the remaining uses of `leftJoinA` over to `optional` in our codebase.

I also appreciate that you're reluctant to make everything `LATERAL` and thus you're unlikely to merge this as is. But it's still want to post this to document the problem, bring the solution up to date against the latest `master` and to start/continue the discussion.

### DIFF
--- a/src/Opaleye/Internal/Optimize.hs
+++ b/src/Opaleye/Internal/Optimize.hs
@@ -19,18 +19,18 @@ optimize = mergeProduct . removeUnit
 removeUnit :: PQ.PrimQuery' a -> PQ.PrimQuery' a
 removeUnit = PQ.foldPrimQuery PQ.primQueryFoldDefault { PQ.product   = product }
   where product pqs = PQ.Product pqs'
-          where pqs' = case NEL.nonEmpty (NEL.filter (not . PQ.isUnit . snd) pqs) of
-                         Nothing -> return (pure PQ.Unit)
+          where pqs' = case NEL.nonEmpty (NEL.filter (not . PQ.isUnit) pqs) of
+                         Nothing -> return PQ.Unit
                          Just xs -> xs
 
 mergeProduct :: PQ.PrimQuery' a -> PQ.PrimQuery' a
 mergeProduct = PQ.foldPrimQuery PQ.primQueryFoldDefault { PQ.product   = product }
   where product pqs pes = PQ.Product pqs' (pes ++ pes')
           where pqs' = pqs >>= queries
-                queries (lat, PQ.Product qs _) = fmap (first (lat <>)) qs
+                queries (PQ.Product qs _) = qs
                 queries q = return q
                 pes' = NEL.toList pqs >>= conds
-                conds (_lat, PQ.Product _ cs) = cs
+                conds (PQ.Product _ cs) = cs
                 conds _ = []
 
 removeEmpty :: PQ.PrimQuery' a -> Maybe (PQ.PrimQuery' b)
@@ -40,9 +40,8 @@ removeEmpty = PQ.foldPrimQuery PQ.PrimQueryFold {
   , PQ.baseTable = return .: PQ.BaseTable
   , PQ.product   = let sequenceOf l = traverseOf l id
                        traverseOf = id
-                       _2 = traverse
                    in
-                   \x y -> PQ.Product <$> sequenceOf (traverse._2) x
+                   \x y -> PQ.Product <$> sequenceOf (traverse) x
                                       <*> pure y
   , PQ.aggregate = fmap . PQ.Aggregate
   , PQ.distinctOnOrderBy = \mDistinctOns -> fmap . PQ.DistinctOnOrderBy mDistinctOns
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -34,17 +34,6 @@ tiToSqlTable ti = HSql.SqlTable { HSql.sqlTableSchemaName = tiSchemaName ti
 
 type Bindings a = [(Symbol, a)]
 
-data Lateral = NonLateral | Lateral
-  deriving Show
-
-instance Semigroup Lateral where
-  NonLateral <> NonLateral = NonLateral
-  _ <> _ = Lateral
-
-instance Monoid Lateral where
-  mappend = (<>)
-  mempty = NonLateral
-
 -- We use a 'NEL.NonEmpty' for Product because otherwise we'd have to check
 -- for emptiness explicity in the SQL generation phase.
 
@@ -56,9 +45,7 @@ instance Monoid Lateral where
 data PrimQuery' a = Unit
                   | Empty     a
                   | BaseTable TableIdentifier (Bindings HPQ.PrimExpr)
-                  | Product   (NEL.NonEmpty (Lateral, PrimQuery' a)) [HPQ.PrimExpr]
-                  -- | The subqueries to take the product of and the
-                  --   restrictions to apply
+                  | Product   (NEL.NonEmpty (PrimQuery' a)) [HPQ.PrimExpr]
                   | Aggregate (Bindings (Maybe (HPQ.AggrOp,
                                                 [HPQ.OrderExpr],
                                                 HPQ.AggrDistinct),
@@ -104,7 +91,7 @@ data PrimQueryFold' a p = PrimQueryFold
   { unit              :: p
   , empty             :: a -> p
   , baseTable         :: TableIdentifier -> Bindings HPQ.PrimExpr -> p
-  , product           :: NEL.NonEmpty (Lateral, p) -> [HPQ.PrimExpr] -> p
+  , product           :: NEL.NonEmpty p -> [HPQ.PrimExpr] -> p
   , aggregate         :: Bindings (Maybe
                              (HPQ.AggrOp, [HPQ.OrderExpr], HPQ.AggrDistinct),
                                    HPQ.Symbol)
@@ -162,7 +149,7 @@ foldPrimQuery f = fix fold
           Unit                        -> unit              f
           Empty a                     -> empty             f a
           BaseTable ti syms           -> baseTable         f ti syms
-          Product qs pes              -> product           f (fmap (fmap self) qs) pes
+          Product qs pes              -> product           f (fmap self qs) pes
           Aggregate aggrs q           -> aggregate         f aggrs (self q)
           DistinctOnOrderBy dxs oxs q -> distinctOnOrderBy f dxs oxs (self q)
           Limit op q                  -> limit             f op (self q)
@@ -178,10 +165,10 @@ foldPrimQuery f = fix fold
         fix g = let x = g x in x
 
 times :: PrimQuery -> PrimQuery -> PrimQuery
-times q q' = Product (pure q NEL.:| [pure q']) []
+times q q' = Product (q NEL.:| [q']) []
 
 restrict :: HPQ.PrimExpr -> PrimQuery -> PrimQuery
-restrict cond primQ = Product (return (pure primQ)) [cond]
+restrict cond primQ = Product (return primQ) [cond]
 
 isUnit :: PrimQuery' a -> Bool
 isUnit Unit = True
--- a/src/Opaleye/Internal/Print.hs
+++ b/src/Opaleye/Internal/Print.hs
@@ -106,7 +106,8 @@ ppSelectLabel l = text "/*" <+> text (preprocess (Sql.lLabel l)) <+> text "*/"
 ppSelectExists :: Exists -> Doc
 ppSelectExists e =
   text "SELECT EXISTS"
-  <+> ppTable (Sql.sqlSymbol (Sql.existsBinding e), Sql.existsTable e)
+  <+> parens (ppSql (Sql.existsTable e)) `ppAs`
+      Just (Sql.sqlSymbol (Sql.existsBinding e))
 
 ppJoinType :: Sql.JoinType -> Doc
 ppJoinType Sql.LeftJoin = text "LEFT OUTER JOIN"
@@ -124,15 +125,11 @@ nameAs :: (HSql.SqlExpr, Maybe HSql.SqlColumn) -> Doc
 nameAs (expr, name) = HPrint.ppSqlExpr expr `ppAs` fmap unColumn name
   where unColumn (HSql.SqlColumn s) = s
 
-ppTables :: [(Sql.Lateral, Select)] -> Doc
+ppTables :: [Select] -> Doc
 ppTables [] = empty
 ppTables ts = text "FROM" <+> HPrint.commaV ppTable_tableAlias (zip [1..] ts)
-  where ppTable_tableAlias :: (Int, (Sql.Lateral, Select)) -> Doc
-        ppTable_tableAlias (i, (lat, select)) =
-          lateral lat $ ppTable (tableAlias i select)
-        lateral = \case
-            Sql.NonLateral -> id
-            Sql.Lateral -> (text "LATERAL" $$)
+  where ppTable_tableAlias :: (Int, Select) -> Doc
+        ppTable_tableAlias (i, select) = ppTable (tableAlias i select)
 
 tableAlias :: Int -> Select -> (TableAlias, Select)
 tableAlias i select = ("T" ++ show i, select)
@@ -142,15 +139,17 @@ ppTable :: (TableAlias, Select) -> Doc
 ppTable (alias, select) = case select of
   Table table           -> HPrint.ppTable table
   RelExpr expr          -> HPrint.ppSqlExpr expr
-  SelectFrom selectFrom -> parens (ppSelectFrom selectFrom)
-  SelectJoin slj        -> parens (ppSelectJoin slj)
-  SelectSemijoin slj    -> parens (ppSelectSemijoin slj)
-  SelectValues slv      -> parens (ppSelectValues slv)
-  SelectBinary slb      -> parens (ppSelectBinary slb)
-  SelectLabel sll       -> parens (ppSelectLabel sll)
-  SelectExists saj      -> parens (ppSelectExists saj)
+  SelectFrom selectFrom -> lateral (ppSelectFrom selectFrom)
+  SelectJoin slj        -> lateral (ppSelectJoin slj)
+  SelectSemijoin slj    -> lateral (ppSelectSemijoin slj)
+  SelectValues slv      -> lateral (ppSelectValues slv)
+  SelectBinary slb      -> lateral (ppSelectBinary slb)
+  SelectLabel sll       -> lateral (ppSelectLabel sll)
+  SelectExists saj      -> lateral (ppSelectExists saj)
   `ppAs`
   Just alias
+  where
+    lateral = (text "LATERAL" $$) . parens
 
 ppGroupBy :: Maybe (NEL.NonEmpty HSql.SqlExpr) -> Doc
 ppGroupBy Nothing   = empty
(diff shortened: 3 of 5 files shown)


duairc:
A sample of the performance improvement this PR brings to some of the more pathological queries:

```diff
-Query/Flow/allOrders                     time                 18.43 s
+Query/Flow/allOrders                     time                 130.8 ms
```

```diff
-Query/Order/countUnassignedOrders        time                 4.483 s
+Query/Order/countUnassignedOrders        time                 41.30 ms
```

```diff
-Query/BoardOrder/findBoardOrderAggregates time                 2.953 s
+Query/BoardOrder/findBoardOrderAggregates time                 2.070 s
```
(note that `findBoardOrderAggregates` returns a massive amount of data, so most of the original `2.070 s` is spent transferring/decoding data, not in the database itself, but it's safe to say that the extra 883 ms caused by the regression *were* spent in the database)

```diff
-Query/PartRank/findPartRankInputs        time                 668.1 s
+Query/PartRank/findPartRankInputs        time                 84.64 s
```

```diff
-Query/Stock/getTraysForFactory           time                 1.567 s
+Query/Stock/getTraysForFactory           time                 799.9 ms
```

duairc:
See [slow.txt](https://github.com/tomjaguarpaw/haskell-opaleye/files/5150237/slow.txt) and [fast.txt](https://github.com/tomjaguarpaw/haskell-opaleye/files/5150238/fast.txt) for the full benchmark outputs. There are a couple of queries that are slower with this change, but not by orders of magnitude,

tomjaguarpaw:
Currently `optional` is as below.  Would it help if I added a `LeftJoinLateral` join type and then changed `join` to `join = PQ.Join PQ.LeftJoin true [] bindings left right`?

```haskell
optional :: SelectArr i a -> SelectArr i (MaybeFields a)
optional = Opaleye.Internal.Lateral.laterally optionalSelect
  where
    -- This is basically a left join on TRUE, but Shane (@duairc)
    -- wrote it to ensure that we don't need an Unpackspec a a.
    optionalSelect :: Select a -> Select (MaybeFields a)
    optionalSelect = IQ.QueryArr . go

    go query ((), left, tag) = (MaybeFields present a, join, Tag.next tag')
      where
        (MaybeFields t a, right, tag') =
          IQ.runSimpleQueryArr (justFields <$> query) ((), tag)

        present = isNotNull (IC.unsafeCoerceColumn t')

        (t', bindings) =
          PM.run (U.runUnpackspec U.unpackspecField (PM.extractAttr "maybe" tag') t)
        join = PQ.Join PQ.LeftJoin true [] bindings left right
    true = HPQ.ConstExpr (HPQ.BoolLit True)
    isNotNull = Opaleye.Internal.Operators.not . Opaleye.Field.isNull
```

tomjaguarpaw:
Or maybe even add a `Lateral` field to the `Join` constructor of `PrimQuery'`.

duairc:
Yes, something like that could work. You probably do just want `LeftJoinLateral` as Postgres complains if you try to make a lateral reference in a full outer join.

duairc:
(`FULL OUTER JOIN LATERAL` syntactically valid, but if you actually try to make a lateral reference it fails)

ocharles:
Out of interest, what is the aversion (maybe that's too strong) to this PR as is?

tomjaguarpaw:
I have a general preference for taking things slowly because if I break something then I might have to do an unknown amount of work to put it back again :)

If we're really certain that using LATERAL everywhere causes no performance loss anywhere then I'd be happy to do that but I'm not sure what evidence could be thoroughly convincing. I know you didn't notice any slow down in your queries but I'm not sure if that's enough to be certain that no one will experience any slow down.

duairc:
That's fair. At this point I'm not fussy about the implementation details, the main thing is just that we can have an `optional` that's performant, and I think a `LeftJoinLateral` constructor would give us that.

tomjaguarpaw:
Nice. There are lots of ways we can tidy up the internals later. For now can you just try adding `LeftJoinLateral` as a join type?

duairc:
@tomjaguarpaw I'm still interested in this by the way, I don't know if you're any more open to it now. I've rebased it onto the latest `master` anyway.

ocharles:
We've just discovered some more motivation for this. In https://github.com/circuithub/rel8/issues/72, it's reported that Rel8 generates a bad plan for a `LEFT JOIN`. What's happening is we're using `optional :: Query a -> Query (MaybeTable a)`, to introduce the `LEFT JOIN`, which works by taking the `LEFT JOIN` of the current `Query` with whatever `Query` is to be made optional. That is, in something like:

```haskell
foo >> optional bar
```

We want something like `Join LeftJoin fooQ barQ`.

This isn't quite what happens at the moment, because `>>=` is implemented in terms of `lateral`, which itself uses `runSimpleQueryArr`. This means what you _actually_ get is:

```haskell
Product [fooQ, Join LeftJoin Unit barQ]
```

This spurious `Unit` wrecks havoc with PostgreSQL's optimizer. First, here's some "ideal" SQL:

```
postgres=# explain select * from (values (1), (2), (3)) s1 (x) left join lateral (select * from (select 0) q, (values (1), (2), (3)) s2 (x) where s1.x = s2.x) s2 on true;
                                QUERY PLAN                                 
---------------------------------------------------------------------------
 Hash Left Join  (cost=0.08..0.15 rows=3 width=12)
   Hash Cond: ("*VALUES*".column1 = "*VALUES*_1".column1)
   ->  Values Scan on "*VALUES*"  (cost=0.00..0.04 rows=3 width=4)
   ->  Hash  (cost=0.04..0.04 rows=3 width=8)
         ->  Values Scan on "*VALUES*_1"  (cost=0.00..0.04 rows=3 width=8)
(5 rows)
```

But look what happens if we introduce a `Unit`:

```
explain select * from (values (1), (2), (3)) s1 (x), lateral (select * from (select 0) q left join lateral (select * from (values (1), (2), (3)) s2 (x) where s1.x = s2.x) s2 on true) q;
                                QUERY PLAN                                 
---------------------------------------------------------------------------
 Nested Loop  (cost=0.00..0.32 rows=3 width=12)
   ->  Values Scan on "*VALUES*"  (cost=0.00..0.04 rows=3 width=4)
   ->  Nested Loop Left Join  (cost=0.00..0.07 rows=1 width=8)
         ->  Result  (cost=0.00..0.01 rows=1 width=4)
         ->  Values Scan on "*VALUES*_1"  (cost=0.00..0.05 rows=1 width=4)
               Filter: ("*VALUES*".column1 = column1)
(6 rows)
```

The worst-case cost has doubled, and we've deteriorated to a loop.

This _could_ be solved in the optimizer, but spotting this unit and re-arranging the query appropriately. I think the simpler option is to just do this PR. Writing optimization passes are really hard and easy to get wrong!

shane-circuithub:
This actually can't be solved in the optimizer alone. This is what such an optimization pass would look like (not sure if 100% correct, but something like this):

```haskell
optional :: Opaleye.PrimQuery' a -> Opaleye.PrimQuery' a
optional = Opaleye.foldPrimQuery Opaleye.primQueryFoldDefault
  { Opaleye.product = product
  }
  where
    product ps es = case go (toList ps) of
      [] -> Opaleye.Unit
      [p] -> snd p
      p : ps' -> Opaleye.Product (p :| ps') es
      where
        go ((x, q) : (x', Opaleye.Join jt on lb rb Opaleye.Unit rq) : rest) =
          (x <> x', Opaleye.Join jt on lb rb q rq) : rest
        go as = as
```

That will basically turn @ocharles's latter example into the former, but it won't be `LEFT JOIN LATERAL`, it will only be `LEFT JOIN`, which will generate an invalid query.

tstat:
AFAIK using a lateral join will force a nested loop (rather than hash join or merge join) unless the postgresql optimizer is able to eliminate the lateral join by [pulling up the lateral subquery](https://github.com/postgres/postgres/blob/04334fde69132f335d9d70cfefe419bd1276b232/src/backend/optimizer/README#L272). The subquery is only able to be pulled up if it satisfies a number of conditions outlined [here](https://github.com/postgres/postgres/blob/04334fde69132f335d9d70cfefe419bd1276b232/src/backend/optimizer/prep/prepjointree.c#L1401). It cannot have any sort of aggregation, limit, offset, window function, etc.

So, to give an example, consider this query:

``` sql
explain
  select
    cl.relname,
    co.conkey
  from pg_class cl
  left outer join pg_constraint co on cl.oid = co.conrelid
  ;
```

which generates the following plan:

``` sql
QUERY PLAN
Hash Left Join  (cost=1.54..22.58 rows=446 width=87)
  Hash Cond: (cl.oid = co.conrelid)
  ->  Seq Scan on pg_class cl  (cost=0.00..17.46 rows=446 width=68)
  ->  Hash  (cost=1.24..1.24 rows=24 width=27)
        ->  Seq Scan on pg_constraint co  (cost=0.00..1.24 rows=24 width=27)
```

We could instead write this with a lateral query

``` sql
explain
  select
    cl.relname,
    t.conkey
  from pg_class cl
  left outer join lateral (
    select *
    from pg_constraint co
    where cl.oid = co.conrelid
  ) as t on true
  ;
```

and generate an identical plan because the subquery satisfies the `is_simple_subquery` predicate.

If we make the subquery slightly more complex in a way that causes it to not satisfy the `is_simple_subquery` predicate then we cannot eliminate the subquery and the lateral remains, which forces the planner to use a nested loop for the lateral join.

Some examples:

``` sql
explain
  select
    cl.relname,
    t.conkey
  from pg_class cl
  left outer join lateral (
    select array_agg(co.conkey) as conkey
    from pg_constraint co
    where cl.oid = co.conrelid
  ) as t on true
  ;
```

``` sql
QUERY PLAN
Nested Loop Left Join  (cost=1.31..613.99 rows=446 width=96)
  ->  Seq Scan on pg_class cl  (cost=0.00..17.46 rows=446 width=68)
  ->  Aggregate  (cost=1.31..1.32 rows=1 width=32)
        ->  Seq Scan on pg_constraint co  (cost=0.00..1.30 rows=2 width=23)
              Filter: (cl.oid = conrelid)
```

compared to a non-lateral equivalent (assuming cl.relname is unique)

``` sql
explain
  select
    cl.relname,
    array_agg(co.conkey)
  from pg_class cl
  left outer join pg_constraint co on cl.oid = co.conrelid
  group by cl.relname
  ;
```

``` sql
QUERY PLAN
HashAggregate  (cost=24.81..30.39 rows=446 width=96)
  Group Key: cl.relname
  ->  Hash Left Join  (cost=1.54..22.58 rows=446 width=87)
        Hash Cond: (cl.oid = co.conrelid)
        ->  Seq Scan on pg_class cl  (cost=0.00..17.46 rows=446 width=68)
        ->  Hash  (cost=1.24..1.24 rows=24 width=27)
              ->  Seq Scan on pg_constraint co  (cost=0.00..1.24 rows=24 width=27)
```

Or we can even force the subquery to not be pulled up with the much sillier `offset 0`

``` sql
explain
  select
    cl.relname,
    t.conkey
  from pg_class cl
  left outer join lateral (
    select *
    from pg_constraint co
    where cl.oid = co.conrelid
    offset 0
  ) as t on true
  ;
```

``` sql
QUERY PLAN
Nested Loop Left Join  (cost=0.00..615.10 rows=892 width=87)
  ->  Seq Scan on pg_class cl  (cost=0.00..17.46 rows=446 width=68)
  ->  Seq Scan on pg_constraint co  (cost=0.00..1.30 rows=2 width=320)
        Filter: (cl.oid = conrelid)
```

The example @ocharles gives above

``` sql
explain
select *
from
  (values (1), (2), (3)) as s1(x)
  cross join lateral (
    select *
    from (select 0) as q
    left join lateral (
      select *
      from (values (1), (2), (3)) as s2(x)
      where s1.x = s2.x
    ) as s2 on true
  ) as q;
```

``` sql
QUERY PLAN
Nested Loop  (cost=0.00..0.29 rows=3 width=12)
  ->  Values Scan on "*VALUES*"  (cost=0.00..0.04 rows=3 width=4)
  ->  Nested Loop Left Join  (cost=0.00..0.07 rows=1 width=8)
        ->  Result  (cost=0.00..0.01 rows=1 width=0)
        ->  Values Scan on "*VALUES*_1"  (cost=0.00..0.05 rows=1 width=4)
              Filter: ("*VALUES*".column1 = column1)
```

fails to satisfy the `is_simple_subquery` predicate for a trickier reason. Have a look at this excerpt

``` c
 * If the subquery is LATERAL, check for pullup restrictions from that.
 */
if (rte->lateral)
{
        bool      restricted;
        Relids        safe_upper_varnos;

        /*
         * The subquery's WHERE and JOIN/ON quals mustn't contain any lateral
         * references to rels outside a higher outer join (including the case
         * where the outer join is within the subquery itself).  In such a
         * case, pulling up would result in a situation where we need to
         * postpone quals from below an outer join to above it, which is
         * probably completely wrong and in any case is a complication that
         * doesn't seem worth addressing at the moment.
         */
        if (lowest_outer_join != NULL)
        {
                restricted = true;
                safe_upper_varnos = get_relids_in_jointree((Node *) lowest_outer_join,
                                                                                                   true);
        }
        else
        {
                restricted = false;
                safe_upper_varnos = NULL; /* doesn't matter */
        }

        if (jointree_contains_lateral_outer_refs((Node *) subquery->jointree,
                                                                                         restricted, safe_upper_varnos))
                return false;
```

I believe that this means that the subquery

``` sql
(
  select *
  from (values (1), (2), (3)) as s2(x)
  where s1.x = s2.x
)
```

fails to be pulled up since we have a qual `s1.x = s2.x` with lateral references to a relation outside of the lowest outer join (the join with `select 0`). It would be wrong to pull this subquery up since that qual would have to be moved from the nullable side of an outer join to the non-nullable side, and would begin filtering rows from the non-nullable side, changing the query.

We then consider pulling up the lateral subquery

``` sql
(
      select *
      from (select 0) as q
      left join lateral (
        select *
        from (values (1), (2), (3)) as s2(x)
        where s1.x = s2.x
      ) as s2 on true
    )
```

which fails for essentially the same reason: If we pulled up this subquery then we cannot have the lateral qual `s1.x = s2.x`, so this condition would need to be pulled up to a where clause, but we refuse to pull it out of the nullable side of an outer join as it would change the meaning of the query. So, we do not pull this subquery up.

So, I believe any sort of query with the structure:

```sql
  select
    ...
  from src
  cross join lateral (
    select
      ...
    from blerg
    left outer join (lateral or not) (
      subquery with some lateral reference to src
    )
  )
```

will fail to have the lateral subquery pulled up, and thus will force a nested loop.

tomjaguarpaw:
I'm going to prioritise this, but rather than make it `LATERAL` all the time (which will likely scare other Opaleye users) I am going to rewrite `lateral` to use `runQueryArr` rather than `runSimpleQueryArr`. That should avoid https://github.com/circuithub/rel8/issues/72.

It needs a bit of fiddling but I hope to have it done this week.

ocharles:
Thank you for that extremely interesting analysis, @tstat! I was not aware of any of that. @tomjaguarpaw your plan sounds like a good one, too - look forward to seeing what you come up with!

tomjaguarpaw:
Interested parties can have a look at https://github.com/tomjaguarpaw/haskell-opaleye/commit/9e6a3138a51b5428b3affff877ccd5ba1648eef5.

It removes the `runSimpleQueryArr` from `bind`, so `bind` never generates units. It is careful to only insert `LATERAL` where it is needed.

It passes the tests but I would welcome anyone trying it on real queries (especially CircuitHub). I'm not completely convinced the tests are thorough enough. (The semi and anti-join bits are almost certainly wrong. I'll fix them later.)

#### Sample code

```haskell
example2 = do
  i <- values [1 :: Field SqlInt4]
  j <- Opaleye.Join.optional (values [i * 10])
  pure (i, j)
```

#### Before

```sql
SELECT
"values0_1" as "result1_4",
NOT (("maybe0_3") IS NULL) as "result2_4",
"values0_2" as "result3_4"
FROM (SELECT
      *
      FROM (SELECT "column1" as "values0_1"
                   FROM
                   (VALUES
                    (CAST(1 AS integer))) as "V") as "T1",
                    LATERAL
                    (SELECT *
                            FROM
                            (SELECT
                             0,
                             *
                             FROM (SELECT
                                   0) as "T1") as "T1"
                                   LEFT OUTER JOIN
                                   (SELECT
                                    CAST(TRUE AS boolean) as "maybe0_3",
                                    *
                                    FROM (SELECT
                                          *
                                          FROM (SELECT "column1" as "values0_2"
                                                       FROM
                                                       (VALUES
                                                        (("values0_1") * (CAST(10 AS integer)))) as "V") as "T1") as "T1") as "T2"
                                                        ON
                                                        TRUE) as "T2") as "T1"
```

#### After

```sql
SELECT
"values0_1" as "result1_4",
NOT (("maybe0_3") IS NULL) as "result2_4",
"values0_2" as "result3_4"
FROM (SELECT *
             FROM
             (SELECT
              *
              FROM (SELECT "column1" as "values0_1"
                           FROM
                           (VALUES
                            (CAST(1 AS integer))) as "V") as "T1") as "T1"
                            LEFT OUTER JOIN
                            LATERAL
                            (SELECT
                             CAST(TRUE AS boolean) as "maybe0_3",
                             *
                             FROM (SELECT
                                   *
                                   FROM (SELECT "column1" as "values0_2"
                                                FROM
                                                (VALUES
                                                 (("values0_1") * (CAST(10 AS integer)))) as "V") as "T1") as "T1") as "T2"
                                                 ON
                                                 TRUE) as "T1"
```

duairc:
I've ported Rel8 to this Opaleye [here](https://github.com/circuithub/rel8/tree/port-opaleye) and I've tried a few of the more pathological queries we have, and they have exactly the same plans as before. Waiting for the full benchmark suite to compile now but it'll looking good.

tomjaguarpaw:
Nice. Thanks for getting that done so quickly. I'm looking forward to hearing the results.

duairc:
Yeah, everything seems to be the same as it was before.

tomjaguarpaw:
Great, I shall do a release containing my proposed changes. (You might want to strengthen the already-released rel8's dependency on Opaleye to something stronger than `<0.8`.)

tomjaguarpaw:
> Great, I shall do a release containing my proposed changes

Published as 0.7.3.0

tomjaguarpaw:
Given that the new `lateral` allows us to make any subquery `LATERAL` without undue performance impact is it OK to close this one?

duairc:
Yes, absolutely, I'm happy with what the current Opaleye has :)

PR #494, opened 2021-03-16
Implement EXISTS directly

tomjaguarpaw:
We already have restrictExists and restrictNotExists, which implement `WHERE EXISTS` and `WHERE NOT EXISTS` respectively, however we can't directly use `EXISTS` to get a `Column PGBool`.

My first attempt at this added `exists` and then redefined `restrictExists` and `restrictNotExists` in terms of `laterally exists`. That basically meant that code that originally looked like this:

```sql
SELECT * FROM
  (SELECT "result0_1"
   FROM
   ...) "T1"
WHERE EXISTS (SELECT "result0_1") as "T2"
```

would now become:

```sql
SELECT * FROM
  (SELECT "result0_1"
   FROM
   ...) "T1",
  LATERAL
  (SELECT EXISTS (SELECT foo
                 ) as "exists0_2") "T1"
WHERE "exists0_2"
```

Unfortunately, in a lot of cases, this made Postgres generate a much worse plan than before. So I've preserved the old constructor in `PrimQuery` (but renamed it to `Semijoin`), and repurposed `Exists` to implement just `EXISTS`.

I also reimplmented `inSelect` in terms of this new `exists` operator, to give a demonstration of its utility.

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -17,6 +17,7 @@ import           Opaleye.Test.TraverseA (traverseA1)
 
 import qualified Opaleye as O
 import qualified Opaleye.Join as OJ
+import qualified Opaleye.Exists as OE
 
 import qualified Database.PostgreSQL.Simple as PGS
 import           Control.Applicative (Applicative, pure, (<$>), (<*>))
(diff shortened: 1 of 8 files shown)


tomjaguarpaw:
Closing in favour of https://github.com/tomjaguarpaw/haskell-opaleye/pull/469#issuecomment-800193641

tomjaguarpaw:
Actually, @duairc suggests this is fine.

tomjaguarpaw:
@duairc and @ocharles can you confirm this is the code we want to merge before I hit the button?

shane-circuithub (approved):
Yes, this looks good to me.

ocharles:
If @shane-circuithub's happy, I'm happy! (I haven't touched this code at all, I just know we rely on it!)

tomjaguarpaw:
Great!

Issue #497, opened 2021-05-03
Support for row_to_json, json_agg, and coalesce to support direct Postgres -> JSON encoding


njaremko:
Does Opaleye support `row_to_json` and `json_agg`? I haven't been able to find it in the docs, but figured I'd ask. If it doesn't, would there be interest in adding it? [Hasura](https://github.com/hasura/graphql-engine) is a haskell project that compiles GraphQL queries into Postgres queries that use the aforementioned functions to have the DB do the JSON encoding (as opposed to querying the DB, parsing the query results, and re-encoding it as JSON) 

An example query looks something like this:
```sql
SELECT
  coalesce(json_agg("root"), '[]') AS "root"
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_4_e"
          FROM
            (
              SELECT
                "_3_root.or.support_ticket_type"."support_ticket_type" AS "support_ticket_type",
                "_0_root.base"."summary" AS "summary",
                "_0_root.base"."details" AS "details"
            ) AS "_4_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public"."support_tickets"
        WHERE
          ('true')
      ) AS "_0_root.base"
      LEFT OUTER JOIN LATERAL (
        SELECT
          row_to_json(
            (
              SELECT
                "_2_e"
              FROM
                (
                  SELECT
                    "_1_root.or.support_ticket_type.base"."id" AS "id",
                    "_1_root.or.support_ticket_type.base"."name" AS "name"
                ) AS "_2_e"
            )
          ) AS "support_ticket_type"
        FROM
          (
            SELECT
              *
            FROM
              "public"."support_ticket_types"
            WHERE
              (
                ("_0_root.base"."support_ticket_type_id") = ("id")
              )
          ) AS "_1_root.or.support_ticket_type.base"
      ) AS "_3_root.or.support_ticket_type" ON ('true')
  ) AS "_5_root"
```

which returns JSON of the form:
```json
[{"support_ticket_type":{"id":1,"name":"....."},"summary":".....","details":"......"}, ...]
```

tomjaguarpaw:
### `json_agg`

[Documentation](https://www.postgresql.org/docs/9.5/functions-aggregate.html)

[...]

PR #498, opened 2021-05-04
Add support for json_agg aggregate function and json_build_object

njaremko:
This PR adds support for the `json_agg` aggregate function, and the `json_build_object` function.

This enables construction of a query that returns it's result as a JSON array bytestring. This can significantly increase throughput since you don't need to query, fetch, parse into records, then serialize into JSON.

```haskell
q = O.aggregate O.jsonAgg $ do
      (firstCol, secondCol) <- O.selectTable table1
      (firstCol2, secondCol2) <- O.selectTable table2
      O.viaLateral O.restrict (firstCol .== firstCol2)
      let blog_post =
            JS.jsonBuildObject $
              JS.jsonBuildObjectField "summary" firstCol2
                <> JS.jsonBuildObjectField "details" secondCol2
      return
        . JS.jsonBuildObject
        $ JS.jsonBuildObjectField "id" firstCol
          <> JS.jsonBuildObjectField "name" secondCol
          <> JS.jsonBuildObjectField "blog_post" blog_post
```
generates this SQL:
```sql
SELECT
"result0_3" as "result1_4"
FROM (SELECT
      *
      FROM (SELECT
            JSON_AGG("inner0_3") as "result0_3"
            FROM (SELECT
                  json_build_object(E'id',"column10_1",E'name',"column21_1",E'blog_post',json_build_object(E'summary',"column10_2",E'details',"column21_2")) as "inner0_3",
                  *
                  FROM (SELECT
                        *
                        FROM (SELECT
                              "column1" as "column10_1",
                              "column2" as "column21_1"
                              FROM "table1" as "T1") as "T1",
                             LATERAL
                             (SELECT
                              "column1" as "column10_2",
                              "column2" as "column21_2"
                              FROM "TABLE2" as "T1") as "T2",
                             LATERAL
                             (SELECT
                              0) as "T3"
                        WHERE (("column10_1") = ("column10_2"))) as "T1") as "T1"
            GROUP BY COALESCE(0)) as "T1") as "T1"
```
which returns this JSON:
```json
[{"id" : 1, "name" : 100, "blog_post" : {"summary" : 1, "details" : 100}}, {"id" : 1, "name" : 100, "blog_post" : {"summary" : 1, "details" : 100}}, {"id" : 1, "name" : 200, "blog_post" : {"summary" : 1, "details" : 100}}]
```

### DIFF
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -143,7 +143,7 @@ boolAnd = A.makeAggr HPQ.AggrBoolAnd
 arrayAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
 arrayAgg = A.makeAggr HPQ.AggrArr
 
-jsonAgg :: Aggregator (C.Column a) (C.Column T.SqlJson)
+jsonAgg :: Aggregator (C.Column a) (C.Column (C.Nullable T.SqlJson))
 jsonAgg = A.makeAggr HPQ.JsonArr
 
 stringAgg :: C.Column T.SqlText
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -143,7 +143,7 @@ boolAnd = A.makeAggr HPQ.AggrBoolAnd
 arrayAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
 arrayAgg = A.makeAggr HPQ.AggrArr
 
-jsonAgg :: Aggregator (C.Column a) (C.Column (C.Nullable T.SqlJson))
+jsonAgg :: Aggregator (C.Column a) (C.Column T.SqlJson)
 jsonAgg = A.makeAggr HPQ.JsonArr
 
 stringAgg :: C.Column T.SqlText
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -25,6 +25,7 @@ module Opaleye.Aggregate
        , boolOr
        , boolAnd
        , arrayAgg
+       , jsonAgg
        , stringAgg
        -- * Counting rows
        , countRows
@@ -142,6 +143,9 @@ boolAnd = A.makeAggr HPQ.AggrBoolAnd
 
 arrayAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
 arrayAgg = A.makeAggr HPQ.AggrArr
+
+jsonAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
+jsonAgg = A.makeAggr HPQ.JsonArr
 
 stringAgg :: C.Column T.SqlText
           -> Aggregator (C.Column T.SqlText) (C.Column T.SqlText)
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -144,7 +144,7 @@ boolAnd = A.makeAggr HPQ.AggrBoolAnd
 arrayAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
 arrayAgg = A.makeAggr HPQ.AggrArr
 
-jsonAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
+jsonAgg :: Aggregator (C.Column a) (C.Column T.SqlJson)
 jsonAgg = A.makeAggr HPQ.JsonArr
 
 stringAgg :: C.Column T.SqlText
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -31,7 +31,6 @@ module Opaleye.Aggregate
        , countRows
        ) where
 
-import           Control.Applicative (pure)
 import           Data.Profunctor     (lmap)
 import qualified Data.Profunctor as P
 
@@ -162,13 +161,16 @@ stringAgg = A.makeAggr' . Just . HPQ.AggrStringAggr . IC.unColumn
 --
 -- See https://github.com/tomjaguarpaw/haskell-opaleye/issues/162
 countRows :: S.Select a -> S.Select (C.Column T.SqlInt8)
-countRows = fmap (C.fromNullable 0)
-            . fmap snd
-            . (\q -> J.leftJoin (pure ())
-                                (aggregate count q)
-                                (const (T.sqlBool True)))
-            . fmap (const (0 :: C.Column T.SqlInt4))
-            --- ^^ The count aggregator requires an input of type
-            -- 'Column a' rather than 'a' (I'm not sure if there's a
-            -- good reason for this).  To deal with that restriction
-            -- we just map a dummy integer value over it.
+countRows =
+  fmap (C.fromNullable 0 . snd)
+    . ( \q ->
+          J.leftJoin
+            (pure ())
+            (aggregate count q)
+            (const (T.sqlBool True))
+      )
+    . fmap (const (0 :: C.Column T.SqlInt4))
+--- ^^ The count aggregator requires an input of type
+-- 'Column a' rather than 'a' (I'm not sure if there's a
+-- good reason for this).  To deal with that restriction
+-- we just map a dummy integer value over it.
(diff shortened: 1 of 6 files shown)


tomjaguarpaw:
I'm not familiar with the use of json_agg. You definitely can't add an alias to a query, but would something like the following work? It should be possible to work out a way of doing that (but perhaps will be equally as hard as `row_to_json`).

```sql
SELECT json_agg(column1, column2)
FROM (SELECT column1, column2 FROM table6)
AS "root"
```

njaremko:
I've added a comment explaining `json_agg` use in more detail in #497

tomjaguarpaw:
Let's discuss back on #497 and revisit this later.

src/Opaleye/Internal/JSONBuildObjectFields.hs
```diff
@@ -0,0 +1,31 @@
...
+      JSONBuildObjectFields $ a <> b
+
+instance Monoid JSONBuildObjectFields where
+  mempty = JSONBuildObjectFields mempty
+
+jsonBuildObjectField :: String -> Column a -> JSONBuildObjectFields
+jsonBuildObjectField f (Column v) = JSONBuildObjectFields [(f, v)]
+
+jsonBuildObject :: JSONBuildObjectFields -> Column SqlJson
+jsonBuildObject (JSONBuildObjectFields jbofs) = Column $ FunExpr "json_build_object" args
```
njaremko, line 28: I ended up unwrapping this here because `concatMap` didn't like `JSONBuildObjectFields` as a `t (String, PrimExpr)` for folding purposes.

tomjaguarpaw: Ah yes, that makes sense.

src/Opaleye/Aggregate.hs
```diff
@@ -143,6 +144,9 @@ boolAnd = A.makeAggr HPQ.AggrBoolAnd
 arrayAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
 arrayAgg = A.makeAggr HPQ.AggrArr
 
+jsonAgg :: Aggregator (C.Column a) (C.Column (T.SqlArray a))
+jsonAgg = A.makeAggr HPQ.JsonArr
```
tomjaguarpaw, line 148: Shouldn't the return type be `Column SqlJson`?

tomjaguarpaw:
Nice! This looks like what I was expecting. Does it work?

src/Opaleye/Internal/HaskellDB/PrimQuery.hs
```diff
@@ -78,7 +78,8 @@ data UnOp = OpNot
 
 data AggrOp     = AggrCount | AggrSum | AggrAvg | AggrMin | AggrMax
                 | AggrStdDev | AggrStdDevP | AggrVar | AggrVarP
-                | AggrBoolOr | AggrBoolAnd | AggrArr | AggrStringAggr PrimExpr
+                | AggrBoolOr | AggrBoolAnd | AggrArr | JsonArr
```
tomjaguarpaw, line 81: CI is complaining because of trailing whitespace on this line

tomjaguarpaw: Looks like you allowed me to push to your branch so I have fixed this.

tomjaguarpaw: Ah, and you did it at the same time :)

njaremko:
So this works:
```haskell
q = O.aggregate O.jsonAgg $ do
      (firstCol, secondCol) <- O.selectTable table6
      return
        . JS.jsonBuildObject
        $ JS.jsonBuildObjectField "summary" firstCol
          <> JS.jsonBuildObjectField "details" secondCol
```

So we're more or less working. All that's left from it being ideal is that I'd like to wrap the result with `COALESCE(q, '[]')`.

tomjaguarpaw:
> All that's left from it being ideal is that I'd like to wrap the result with `COALESCE(q, '[]')`.

Ah, does aggregating an empty group return `NULL`? We probably ought to discuss that further. I'm pretty negative on aggregates of empty queries producing *any* rows. See the note at https://hackage.haskell.org/package/opaleye-0.7.1.0/docs/Opaleye-Aggregate.html#v:aggregate

(Perhaps the same is true of `array_agg`. If so I shall have to fix that too.)

njaremko:
Yep, the following returns NULL:
```sql
SELECT
  json_agg("root")
FROM 
  (SELECT column1 as "name", column2 as "value" FROM table6 WHERE column1 = '23') AS "root"
```
To be clear, I don't want to make `jsonAgg` coalesce by default, but I would like to have the option to wrap it in a coalesce.

tomjaguarpaw:
> I don't want to make `jsonAgg` coalesce by default, but I would like to have the option to wrap it in a coalesce.

You can just call [`fromNullable`](https://hackage.haskell.org/package/opaleye-0.7.1.0/docs/Opaleye-Column.html#v:fromNullable) on it (which isn't implemented in terms of `COALESCE` but could be, I think). There's no need to have `COALESCE` in the aggregator itself..

tomjaguarpaw:
Would you be OK with `jsonAgg` of an empty query returning an empty query?

njaremko:
So change `jsonAgg` to be nullable, and fromNullable should handle it fine?
```haskell
jsonAgg :: Aggregator (C.Column a) (C.Column (C.Nullable T.SqlJson))
jsonAgg = A.makeAggr HPQ.JsonArr
```

tomjaguarpaw:
By empty query I mean one that has no rows, for example the subquery in your [example above](https://github.com/tomjaguarpaw/haskell-opaleye/pull/498#issuecomment-833740247) or the subquery in the below. Postgres returns `NULL` for these queries but that doesn't make any sense mathematically. The aggregation of an empty group should be empty, even if there is only one group.

Opaleye has some logic to make aggregations of empty groups return no rows.

The question for you is whether you want Opaleye's behaviour or you really do want to get an empty list when aggregating an empty query.

```sql
SELECT json_agg("root")
FROM (SELECT 0 WHERE FALSE) AS "root"
```

tomjaguarpaw:
> So change` jsonAgg` to be `Nullable`, and `fromNullable` should handle it fine?

I think this is actually made moot by my observation above. The type should be

```
jsonAgg :: Aggregator (C.Column a) (C.Column T.SqlJson)
```

and it *can't* return `NULL`.

(Caveat: We might have to use a more specific type than `a`. I'm not sure what yet.)

src/Opaleye/Internal/JSONBuildObjectFields.hs
```diff
@@ -0,0 +1,35 @@
...
+where
+
+import Opaleye.Internal.Column (Column (Column))
+import Opaleye.Internal.HaskellDB.PrimQuery (Literal (StringLit), PrimExpr (ConstExpr, FunExpr))
+import Opaleye.Internal.PGTypesExternal (SqlJson)
+
+newtype JSONBuildObjectFields
+  = JSONBuildObjectFields [(String, PrimExpr)]
+
+instance Semigroup JSONBuildObjectFields where
```
tomjaguarpaw, line 16: You'll need to import `Data.Semigroup` explicitly so that it works on GHC 8.0 and 8.2.

njaremko:
If I remove `Nullable` from the type signature, as you've described above. What happens if I tell Opaleye to give me an aggregation on a query that returns no rows?

tomjaguarpaw:
It will return a query with no rows.

njaremko:
Played with it a bit, and I get it now. I've removed `Nullable` from the type signature of `jsonAgg`.

Do you have a better module for me to move this new json stuff into? Or do you want to make a new module?

tomjaguarpaw:
Nice! If you are confident that it works for your purposes then I will make some administrative changes locally regarding modules and API and then push for you to review.

njaremko:
Administrate to your hearts content 👍

njaremko:
You can ignore these comments, I realized a few of my comments were "pending" as part of a github review that I hadn't clicked submit on.

tomjaguarpaw:
This is looking great! Closing this one in favour of https://github.com/tomjaguarpaw/haskell-opaleye/pull/500.

Issue #508, opened 2021-05-26
Adding `now()` function support


njaremko:
Would a PR adding support for the `now()` function be welcomed? Maybe it's just me, but I use it frequently, so it's absence from opaleye has been a pain in the butt :)

tomjaguarpaw:
Sure, go for it!

basvandijk:
I see there already is a [now](https://hackage.haskell.org/package/opaleye-0.7.3.0/docs/Opaleye-Operators.html#v:now) function.

tomjaguarpaw:
Yes, we can close this. Thanks for noticing.

PR #511, opened 2021-06-11
Add now function support

njaremko:
This addresses #508 . Sorry this took a while, I got side tracked on another project. 

This PR adds support for the `now()` postgres function: https://www.postgresql.org/docs/current/functions-datetime.html

Wasn't really sure where to put it, so I threw it in `Operators`. If you want it somewhere else, just let me know.


tomjaguarpaw:
Fantastic, thank you!

PR #517, opened 2021-07-11
Fix typo in turorial

cmcaine:
Hi!

I'm not a haskeller, but this looks right to me. Sorry if this is noise!

### DIFF
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -835,7 +835,7 @@ We could run the select `selectTable goodWarehouseTable` like this.
 > type GoodWarehouse = Warehouse' WarehouseId String Int
 >
 > runWarehouseSelect :: PGS.Connection
->                   -> Select GoodWarehouseField
+>                   -> Select GoodWarehouse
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Hi, thanks for the interest in Opaleye! As you can see from [the test results](https://github.com/tomjaguarpaw/haskell-opaleye/pull/517/checks?check_run_id=3039537008#step:8:772) the original formulation was correct. On the SQL side we have a `Select` of "fields" and it turns into a normal Haskell value when you run the `Select`. It's no problem to have raised the issue though.

If you're not a Haskeller I'm curious what brought you here. Was it my post on HN?

cmcaine:
Aye, that does make sense, probably shouldn't try to think before breakfast :)

Someone linked opaleye in this thread: https://lobste.rs/s/jxeg1b/against_sql mentioning it as a good SQL library, so I just thought I'd take a look and it does seem good from what I did understand!

All the best,

tomjaguarpaw:
Ah, interesting. I hadn't seen that. Thanks!

You might also be interested in https://rel8.readthedocs.io/en/latest/ which is a wrapper on top of Opaleye that is more user friendly in some ways.

PR #518, opened 2021-07-15
distinctOn an empty set of columns should be `limit 1`, not a no-op

duairc:
I understand why this is made a no-op, because an empty list of columns in a `DISTINCT ON` query is not syntactically valid in SQL. However, our Haskell types don't forbid it. So the question is, semantically, what does this mean? And I think it has to be an unconditional `limit 1`.

### DIFF
--- a/src/Opaleye/Internal/Order.hs
+++ b/src/Opaleye/Internal/Order.hs
@@ -76,6 +76,14 @@ distinctOnBy ups proj ord (cols, pq, t) = (cols, pqOut, t)
             x:xs -> PQ.DistinctOnOrderBy (Just $ x NL.:| xs) (orderExprs cols ord) pq
             []   -> pq
 
+distinctOnByCorrect :: U.Unpackspec b b -> (a -> b) -> Order a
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Oh yes, I think you're right! It will require a new major release. How awkward. Maybe that will prompt me to finally finish the `Column` -> `Field` migration.

tomjaguarpaw:
Would you like to add this as a *new* function `distinctOnByCorrect`? Then at least it will be in Opaleye and rel8 can use it without having to implement its own version. Then in version 8 I will replace `distinctOnBy` with it and deprecate `distinctOnByCorrect`.

(I'll get round to this sometime in the near future unless you do.)

duairc:
Sure. I just force-pushed that change.

tomjaguarpaw:
I've implemented this and will do a release at some point in the near future.

tomjaguarpaw:
Released as https://hackage.haskell.org/package/opaleye-0.7.4.0

PR #519, opened 2021-07-20
Add support for interval types and + and - operations

### LINKED ISSUE #509: How to add intervals to dates?
I was trying to figure out how to add intervals to the SqlDate or SqlTimestamp types, but I couldn't find any interval types, or anything in Opaleye.Operators. Is this implemented anywhere?

basvandijk:
This adds the `SqlInterval` type matching PostgreSQL's `interval` type.

`SqlInterval` can be converted to Haskell's `CalendarDiffTime` type.

The operations `addInterval` and `minusInterval` have been added:

```
addInterval, minusInterval
  :: IntervalNum from to
  => F.Field from
  -> F.Field T.SqlInterval
  -> F.Field to
```

They add or subtract an interval from various date and time types. Which types are allowed to add or subtract an interval from and which type is the resulting type is determined by the:

```
class IntervalNum from to

instance IntervalNum T.SqlDate        T.SqlTimestamp
instance IntervalNum T.SqlInterval    T.SqlInterval
instance IntervalNum T.SqlTimestamp   T.SqlTimestamp
instance IntervalNum T.SqlTimestamptz T.SqlTimestamptz
instance IntervalNum T.SqlTime        T.SqlTime
```

Simple test cases have been added to test these operations.

This fixes #509.

### DIFF
--- a/src/Opaleye/Internal/PGTypesExternal.hs
+++ b/src/Opaleye/Internal/PGTypesExternal.hs
@@ -107,8 +107,8 @@ pgTimeOfDay = IPT.unsafePgFormatTime "time" "'%T%Q'"
 -- "We recommend not using the type time with time zone"
 -- http://www.postgresql.org/docs/8.3/static/datatype-datetime.html
 
-pgInterval :: Time.CalendarDiffTime -> Column PGInterval
-pgInterval = IPT.castToType "interval" . quote . Time.Format.ISO8601.iso8601Show
+sqlInterval :: Time.CalendarDiffTime -> Column PGInterval
+sqlInterval = IPT.castToType "interval" . quote . Time.Format.ISO8601.iso8601Show
     where
       quote s = "'" ++ s ++ "'"
 
--- a/src/Opaleye/Internal/PGTypesExternal.hs
+++ b/src/Opaleye/Internal/PGTypesExternal.hs
@@ -20,7 +20,8 @@ import qualified Data.Text.Lazy as LText
 import qualified Data.ByteString as SByteString
 import qualified Data.ByteString.Lazy as LByteString
 import           Data.Scientific as Sci
-import qualified Data.Time as Time
+import qualified Data.Time.Compat as Time
+import qualified Data.Time.Format.ISO8601.Compat as Time.Format.ISO8601
 import qualified Data.UUID as UUID
 
 import           Data.Int (Int64)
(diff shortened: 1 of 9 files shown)


src/Opaleye/Operators.hs
```diff
@@ -469,6 +473,27 @@ timestamptzAtTimeZone = C.binOp HPQ.OpAtTimeZone
...
 dateOfTimestamp (Column e) = Column (HPQ.FunExpr "date" [e])
 
+-- | @IntervalNum from to@ determines from which date or time types an interval
+-- can be added ('addInterval') or subtracted ('minusInterval`) and which is the
+-- resulting type.
+--
+-- The instances should correspond to the interval + and - operations listed in:
+--
+-- https://www.postgresql.org/docs/current/functions-datetime.html#OPERATORS-DATETIME-TABLE
+class IntervalNum from to
```
tomjaguarpaw, line 483: Would it be sensible to have a functional dependency here?

src/Opaleye/Internal/PGTypesExternal.hs
```diff
@@ -106,6 +107,10 @@ pgTimeOfDay = IPT.unsafePgFormatTime "time" "'%T%Q'"
 -- "We recommend not using the type time with time zone"
 -- http://www.postgresql.org/docs/8.3/static/datatype-datetime.html
 
+pgInterval :: Time.CalendarDiffTime -> Column PGInterval
+pgInterval = IPT.castToType "interval" . quote . Time.Format.ISO8601.iso8601Show
+    where
+      quote s = "'" ++ s ++ "'"
```
tomjaguarpaw, line 113: Probably best just to call this `sqlInterval`. The `pg...` naming is deprecated, and we don't need to continue to use it, even in internal modules.

src/Opaleye/SqlTypes.hs
```diff
@@ -170,6 +173,8 @@ sqlTimeOfDay = P.pgTimeOfDay
 -- "We recommend not using the type time with time zone"
 -- http://www.postgresql.org/docs/8.3/static/datatype-datetime.html
 
+sqlInterval :: Time.CalendarDiffTime -> F.Field SqlInterval
+sqlInterval = P.pgInterval
```
tomjaguarpaw, line 177: Then we can just re-export `sqlInterval` rather than making up a new name for it.

tomjaguarpaw:
Looks great, thanks! Just a couple of comments.

src/Opaleye/Operators.hs
basvandijk: Initially I thought we could have arbitrary combinations but thinking about it again I think it makes sense to add a fundep here. Will do.

basvandijk:
Thanks for the review @tomjaguarpaw. I agree with all your comments and I've addressed them. Could you do another review?

tomjaguarpaw:
This is great, thank you @basvandijk!

PR #524, opened 2021-09-05
WIP matchMaybe

tomjaguarpaw:
This looks super useful. I'm not sure why we didn't come up with this before!

Any thoughts? (Particularly @ocharles and @duairc)


ocharles:
Oh, this is pretty cool. I have a little prototype of something similar but wilder - I made `Query` have `Cont` like capabilities, then 

```haskell
x <- leftJoin ...
... x
```

has `leftJoin` capture the continuation, and then invoke the continuation with `x = Nothing` and `x = Just x'` and `union` the results. I was a bit nervous of that though, because obviously it incurs some pretty significant SQL. This looks to be basically the same idea, but rather than doing it at the `Query` level, you're doing it more at the expression level. It's also much more opt-in than my approach.

I'm not sure about the type class and the DSL, I think `matchMaybe` alone is fine. With `-XBlockArguments`, you get:

```haskell
foo <- leftJoin ...
matchMaybe foo \case
  Nothing -> bar
  Just x -> baz x
```

Which looks perfectly reasonable to me.

Something else I've been considering oddly just got published to the Arxiv, which is to use pattern synonyms. https://arxiv.org/abs/2108.13114 is the paper in question, but the idea is essentially to have:

```haskell
pattern Just_
pattern Nothing_
```

and to have these transparently look at the tag.

So there's a whole bunch of stuff here, all with merit. I dunno which to pick!

tomjaguarpaw:
> I'm not sure about the type class and the DSL, I think matchMaybe alone is fine. 

Yeah, that's fair.

> Something else I've been considering oddly just got published to the Arxiv

Interesting! The [Reddit discussion](https://www.reddit.com/r/haskell/comments/phx4me/embedded_pattern_matching/.compact) prompted me to consider this but I haven't actually read the paper yet.

PR #527, opened 2021-09-11
add `PGVarcharN`

stevemao:
I think it makes sense to use `PGString` where possible.


tomjaguarpaw:
Hi, sorry I haven't given this a thorough review yet. I suspect for an early design it would be simplest to add only `SqlVarcharN` and conversions to and from `SqlText`. Hopefully I'll be able to get back to this in the near future.

tomjaguarpaw:
OK, so for now I am considering adding only `SqlVarcharN` and conversions to and from `SqlText`. Would that work for you? Extending `(.++)`, `stringAgg` and so on doesn't seem very useful as the `VARCHAR` is likely to overflow. What do you think?

PR #530, opened 2021-09-26
Postgres treats upper case letters in type names as lower case,

stevemao:
unless the name is quoted!

### DIFF
--- a/src/Opaleye/Experimental/Enum.hs
+++ b/src/Opaleye/Experimental/Enum.hs
@@ -5,6 +5,7 @@ module Opaleye.Experimental.Enum
   (
     enumMapper,
     EnumMapper,
+    enumMapperWithSchema,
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Ah, good catch, thank you. I'm rather backlogged at the moment but I'll get to this as soon as I can.

tomjaguarpaw:
Cool, thank you. Please let me know if you need me to make a release containing this.

stevemao:
a new release would be nice. So I can remove my workaround 😄

tomjaguarpaw:
OK done! https://hackage.haskell.org/package/opaleye-0.7.5.0

stevemao:
@tomjaguarpaw do you mind pushing the commit and the tag? incase the version confuses people

PR #531, opened 2021-10-16
add sqlEnum to construct Column enum 

stevemao:
I need to filter on the enum field like this

```
where_ $ sqlRatingType row .== sqlRating PG
```


tomjaguarpaw:
I still don't understand why `toFields` isn't sufficient. If you want to avoid the `Default` constraint then maybe add `toFieldsExplicit :: ToFields h f -> h -> f` to `Opaleye.ToFields`?

stevemao:
My bad. Sorry about this.

tomjaguarpaw:
That's OK!

PR #551, opened 2022-06-11
Implement WITH RECURSIVE

shane-circuithub:
This is based on @hesselink's original attempt in #58, but we use a `Binaryspec` instead of an `Unpackspec` (to ensure that both the base and recursive cases have the same number of columns).

Also, I've simplified the `PrimQuery` constructor; it's no longer specific to `WITH RECURSIVE`; it could easily be used for ordinary `WITH` expressions as well.

In terms of the implementation, I use a `LATERAL` product to avoid the need for the `addTable` function that the original implementation had. I don't know if there are performance implications for this in general, but for the one query I've tested this on it produces the same plan whether using `LATERAL` or  the `addTable` method.


src/Opaleye/WithRecursive.hs
```diff
@@ -0,0 +1,78 @@
...
+import Opaleye.Internal.Unpackspec (Unpackspec(..), runUnpackspec)
+import Opaleye.Internal.PackMap (PackMap(..))
+import qualified Opaleye.Internal.HaskellDB.PrimQuery as HPQ
+import qualified Opaleye.Internal.PrimQuery           as PQ
+import qualified Opaleye.Internal.Sql                 as Sql
+import qualified Opaleye.Internal.Tag                 as Tag
+import qualified Opaleye.Internal.PackMap as PM
+
+
+withRecursive :: Default Binaryspec a a => Select a -> SelectArr a a -> Select a
```
tomjaguarpaw, line 22: Do you know if this can be generalised to `Select a -> (a -> Select a) -> Select a`? If it can it would be good to do so. If it can't it would be good to document why.

Given that there's a `LATERAL` perhaps it can. I'd also like to know to what extent it depends on the `LATERAL`, perhaps having another version without the `LATERAL` that uses `addTable` (although I don't understand `addTable`).

tomjaguarpaw:
Fabulous, thanks! I'll have to give it a deeper look when I have more time, but it looks great to me.

src/Opaleye/WithRecursive.hs
shane-circuithub: You're right that the type can be generalised to `Select a -> (a -> Select) -> Select a`, so I've done that.

shane-circuithub: As for another version that uses `addTable`, I don't fully understand `addTable` either, and the comments on the previous implementation implied that it was incomplete. As for the extent to which it uses `LATERAL`, maybe I'll just show the SQL it generates:

```sql
SELECT
CAST("with0_8" AS int8) as "anon"
FROM (SELECT
      *
      FROM (WITH RECURSIVE "cte_1" AS
            (SELECT
             "values0_2" as "binary0_3"
             FROM (SELECT
                   *
                   FROM (SELECT "column1" as "values0_2"
                         FROM
                         (VALUES
                          (CAST(1 AS int8)),
                          (CAST(2 AS int8)),
                          (CAST(3 AS int8)),
                          (CAST(5 AS int8))) as "V") as "T1") as "T1"
             UNION ALL
             SELECT
             "values0_5" as "binary0_3"
             FROM (SELECT
                   *
                   FROM (SELECT
                         "binary0_3" as "rec0_4"
                         FROM "cte_1" as "T1") as "T1",
                        LATERAL
                        (SELECT "column1" as "values0_5"
                         FROM
                         (VALUES
                          (("rec0_4") * (CAST(2 AS int8))),
                          (("rec0_4") * (CAST(3 AS int8)))) as "V") as "T2"
                   WHERE (("values0_5") < (CAST(10 AS int8)))) as "T1")
            SELECT
            "binary0_3" as "with0_8"
            FROM "cte_1" as "T1") as "T1") as "T1"
```

(That's generated by the Rel8 code:
```haskell
putStrLn $ showQuery $ loop (values [1 :: Expr Data.Int.Int64, 2, 3, 5]) (\a -> do x <- values [a * 2, a * 3]; where_ (x <. 10); pure x)
```
but the Opaleye would be very similar)

So, the general form is:

```sql
WITH RECURSIVE "cte_N" AS
      (SELECT
       [...] as "binaryN_N"
       FROM (...) as "T1" -- base case
       UNION ALL
       SELECT
       [...] as "binaryN_N"
       FROM (SELECT
             *
             FROM (SELECT
                   "binaryN_N" as "recN_N"
                   FROM "cte_1" as "T1") as "T1",
                  LATERAL
                  (...) as "T2") as "T1") -- recursive case, from inside which
                                          -- we want to be able to refer to
                                          -- columns like recN_N
      SELECT
      "binaryN_N" as "withN_N"
      FROM "cte_N" as "T1"
```

tomjaguarpaw:
I'm happy to add the `LATERAL`-using version and consider a non-`LATERAL` version later if people want it.

I still haven't had time to look at this properly. I'd particularly like to add some property tests for it. If the delay in getting this merged is an inconvenience then please let me know and I can prioritise it.

src/Opaleye.hs
```diff
@@ -36,6 +36,8 @@ module Opaleye ( module Opaleye.Adaptors
                , module Opaleye.Table
                , module Opaleye.ToFields
                , module Opaleye.Values
+               , module Opaleye.Window
```
tomjaguarpaw, line 39:
```suggestion
```

src/Opaleye.hs
```diff
@@ -61,3 +63,5 @@ import Opaleye.SqlTypes
 import Opaleye.Table
 import Opaleye.ToFields
 import Opaleye.Values
+import Opaleye.Window
```
tomjaguarpaw, line 66:
```suggestion
```

src/Opaleye/With.hs
```diff
@@ -0,0 +1,108 @@
...
+  let
+    mkSelect = PQ.BaseTable (PQ.TableIdentifier Nothing (Sql.sqlSymbol tableName))
+
+  (a, primQBase) <- runSimpleQueryArr' base ()
+
+  startTag <- Tag.fresh
+
+  let
+    (columns, _) = PM.run $
+      runUnpackspec unpackspec (PM.extractAttr "binary" startTag) a
```
tomjaguarpaw, line 70:
```suggestion
      runUnpackspec unpackspec (PM.extractAttr "base" startTag) a
```

Perhaps a better name?

src/Opaleye/With.hs
```diff
@@ -0,0 +1,108 @@
...
+import qualified Opaleye.Internal.Tag                 as Tag
+import qualified Opaleye.Internal.PackMap as PM
+
+
+with :: Default Unpackspec a a => Select a -> (Select a -> Select b) -> Select b
+with = withExplicit def
+
+
+withRecursive :: Default Binaryspec a a => Select a -> (a -> Select a) -> Select a
+withRecursive = withRecursiveExplicit def
```
tomjaguarpaw, line 29:
```suggestion
-- | @withRecursive s f@ is the smallest set of rows @r@ such that
--
-- @
-- r == s \`'unionAll'\` (r >>= f)
-- @
withRecursive :: Default Binaryspec a a => Select a -> (a -> Select a) -> Select a
withRecursive = withRecursiveExplicit def
```

Is this correct? If so, shall we include it?

tomjaguarpaw:
This looks great., thanks! Would you mind addressing the comments I've made? Then I will be happy to merge it.

src/Opaleye/With.hs
shane-circuithub: I'm pretty sure it needs to be `binary` because the column names from [here](https://github.com/circuithub/haskell-opaleye/blob/214e2841f03a6ecf3d64138985fb3dc6df787488/src/Opaleye/With.hs#L75) need to match the column names from [here](https://github.com/circuithub/haskell-opaleye/blob/214e2841f03a6ecf3d64138985fb3dc6df787488/src/Opaleye/With.hs#L90).

shane-circuithub:
I've addressed your comments :)

src/Opaleye/With.hs
tomjaguarpaw: Hmm, there's something funny going on because `extractAttr` ignores the `PrimExpr`s inside the `a` argument to `runUnpackspec`. That suggests that if you delete

```haskell
  let
    (columns, _) = PM.run $
      runUnpackspec unpackspec (PM.extractAttr "binary" startTag) a
```

and use `a` instead of `columns` where the latter appears then you should get the same result. I'll dig into this more when I get some time (hopefully at the weekend) unless you want to investigate before then.

tomjaguarpaw: If tried it, and it's not possible to use just `columns = a`. I guess because of the recursive nature of the thing we have to get `columns` before the thing that morally produces columns is run. I don't really quite grasp what's going on here, but I'm pretty confident that it works fine, so hopefully I'll be able to merge soon.

tomjaguarpaw:
Now that I understand `WITH` a bit better I was able to extract out some of the functionality. @shane-circuithub, I will merge my version in due course and I would welcome any comments from you if you have time to look at it.

https://github.com/tomjaguarpaw/haskell-opaleye/pull/554/files#diff-376686cd65748aaa210ad0582504b4cdd3983cc11348bcdf313491d09e43841d

tomjaguarpaw:
Superceded by https://github.com/tomjaguarpaw/haskell-opaleye/commit/6cc38b455e1250d33e2a1e1b85086710a0c25cee

tomjaguarpaw:
Thanks to @shane-circuithub and @hesselink for your work on this! Great to finally have it in Opaleye.

PR #552, opened 2022-06-19
Add support for window functions

### LINKED ISSUE #149: Window functions
Hi,

Any plans to be able to use window functions in querying with opaleye (or can you already)?

http://www.postgresql.org/docs/9.5/static/functions-window.html

shane-circuithub:
This would fix #149.

I haven't written any documentation for this yet, but basically you have the function `window :: Select (Window a) -> Select a`, where `Window` is an `Applicative`. To construct a `Window` value, you can use `pure` (which has no effect), the named window functions we support (e.g., `rowNumber :: Field SqlInt8`), or `cumulative`, which takes any `Aggregator` and makes a window function version (e.g., `cumulative sum :: Field a -> Window (Field a)`). To specify a partition, you can use the `over` function `Window a -> Partition -> Window a`, where `Partition` is constructed by `partitionBy` or `orderBy`. `Partition` is a `Monoid` so these can be trivially combined together.

The only open question I have really is about the name `orderBy`. Obviously we couldn't re-export this from `Opaleye` as is because the name `orderBy` is already taken. We could call it something else, or just tell users to explicitly import `Opaleye.Window`, but it's up to you.


tomjaguarpaw:
Cool, thank you!  Again I'll need a bit of time to look at this. Please nudge me if you need me to prioritise getting this merged.

shane-circuithub:
Hey, I added documentation to all the functions introduced here. I also just renamed `orderBy` to `orderParititonBy` to resolve the naming conflict. Obviously take your time to review it as you see fit, but I think it's probably in a good state now.

tomjaguarpaw:
Very nice, thanks!  And thanks for the great documentation too.

My main question is whether this API is safe in the absence of `LATERAL`. I would guess not.  I guess a safe API would be closer to the `Aggregator` API, something like:

```haskell
window :: Select a -> Window a b -> Select b

over :: Window a b -> Partition a -> Window a b

orderPartitionBy :: Order a -> Partition a

lag :: Window (Field_ n a, Field SqlInt4, Field_ n a) (Field_ n a) 
```

etc. with `Window` a `ProductProfunctor` and `Partition` a "product contravariant" a.k.a. `Divisible`.

But I'm by no means certain about my scepticism regarding safety. I also suspect that being safe in this regard will compromise the nicer API in the presence of `LATERAL` that you probably want for `rel8`. Is that right?

If so then perhaps we can come to the same arrangement as we do for the `Aggregator` API, where `rel8` uses some of the internals rather than the public API.

What do you think?

src/Opaleye/Internal/Window.hs
```diff
@@ -0,0 +1,119 @@
...
+  pm (\(mop, expr) -> case mop of
+         Nothing -> pure expr
+         Just (op, _, _) -> f (HPQ.WndwAggregate op expr, mempty)) a
+
+
+-- | 'over' adds a 'Partition' to a 'Window' expression.
+--
+-- @@@
+-- 'cumulative' 'Opaleye.sum' salary `over` 'partitionBy' department <> 'orderPartitionBy' salary ('Opaleye.desc' id)
+-- @@@
```
tomjaguarpaw, line 84:
```suggestion
-- | 'over' adds a 'Partition' to a 'Window' expression.
--
-- @
-- 'cumulative' 'Opaleye.sum' salary \`'over'\` 'partitionBy' department <> 'orderPartitionBy' salary ('Opaleye.desc' id)
-- @
```

This seems to be required to get the Haddock to render as I anticipated locally. Do you agree?

tomjaguarpaw:
I'm still slowly working my way through understanding window functions. As ever please ping me if you need me to prioritise this.

tomjaguarpaw:
As predicted, I was more comfortable with a `Profunctor`-based API, which I have proposed at https://github.com/tomjaguarpaw/haskell-opaleye/pull/559, and intend to move ahead with that version in due course. Of course, you can use some of the internals to expose your preferred API style for rel8.

PR #561, opened 2023-02-09
Basic Full Text Search support

CGenie:
Hello,

I have this code that we added some time ago to support basic Full Text Search queries in Postgres. 

https://www.postgresql.org/docs/current/textsearch.html

It basically contains definitions of `tsvector` and `tsquery` types and the `@@` operator for matching. We use this for quite some time now and it seems to work. I would much prefer to use upstream opaleye than the patched version. :)


opaleye.cabal
```diff
@@ -39,6 +39,7 @@ library
     , scientific          >= 0.3     && < 0.4
     , semigroups          >= 0.13    && < 0.21
     , text                >= 0.11    && < 2.1
+    , time
```
tomjaguarpaw, line 42: I don't understand why `time` is needed.

src/Opaleye/Internal/PGTypesExternal.hs
```diff
@@ -182,6 +182,15 @@ pgRange pgEl start end =
         oneEl R.NegInfinity   = HPQ.NegInfinity
         oneEl R.PosInfinity   = HPQ.PosInfinity
 
+-- Full Text Search
+
+pgTSVector :: String -> Field PGTSVector
+pgTSVector = IPT.castToType "tsvector" . HSD.quote
+
+pgTSQuery :: String -> Field PGTSQuery
+pgTSQuery = IPT.castToType "tsquery" . HSD.quote
```
tomjaguarpaw, line 191: Shouldn't these take `Field PGText` as argument, rather than `String`? Otherwise you can't do text search on anything that's actually available in your database.

src/Opaleye/Internal/PGTypesExternal.hs
```diff
@@ -308,3 +322,5 @@ type PGBytea = SqlBytea
 type PGJson = SqlJson
 type PGJsonb = SqlJsonb
 type PGRange = SqlRange
+type PGTSQuery = SqlTSQuery
+type PGTSVector = SqlTSVector
```
tomjaguarpaw, line 326: Existing types named `PG*` are deprecated, so please don't add new ones.

```suggestion
```

src/Opaleye/Operators.hs
```diff
@@ -212,6 +213,12 @@ infix 4 .>=
 (.>=) :: Ord.SqlOrd a => Field a -> Field a -> F.Field T.SqlBool
 (.>=) = C.binOp (HPQ.:>=)
 
+infix 4 @@
+(@@) :: Field a -> Field T.SqlTSQuery -> F.Field T.SqlBool
+(@@) = C.binOp (HPQ.:@@)
```
tomjaguarpaw, line 218: Shouldn't this be well-typed?

```suggestion
(@@) :: Field T.SqlTSVector -> Field T.SqlTSQuery -> F.Field T.SqlBool
(@@) = C.binOp (HPQ.:@@)
```

tomjaguarpaw (commented):
Thanks! I have some questions about things I didn't understand.

opaleye.cabal
CGenie: Removed.

src/Opaleye/Internal/PGTypesExternal.hs
CGenie: Yes, that's a good idea, thanks.

src/Opaleye/Internal/PGTypesExternal.hs
CGenie: OK, removed, but how does this compare to the remark above to use `Field PGText`? Should that be `Field SqlText` instead?

tomjaguarpaw: Oh, yes, I should have said `Field SqlText` above.

src/Opaleye/Internal/PGTypesExternal.hs
CGenie: Ah, ok, but a question: how should I proceed with destructuring `PGText`? I found `unColumn` but it's not enough. When constructing the `String -> Field PGTSVector` type, I used as example the `pgJSON` etc functions in the same file.

tomjaguarpaw: I guess `FunExpr`, for example https://github.com/tomjaguarpaw/haskell-opaleye/blob/912641cddd98bbdd315f5614b4098dc2460064ee/src/Opaleye/Operators.hs#L291-L292

src/Opaleye/Internal/PGTypesExternal.hs
CGenie: Is this OK now?

src/Opaleye/Operators.hs
CGenie: Is this OK now?

src/Opaleye/Internal/PGTypesExternal.hs
tomjaguarpaw: Looks fine, thanks.

src/Opaleye/Operators.hs
tomjaguarpaw: Looks fine, thanks.

tomjaguarpaw:
Could you please rebase this as a single commit on `master`?

PR #565, opened 2023-05-06
Add support for Postgres' `FILTER (WHERE ...)` syntax in aggregations

tomjaguarpaw:


shane-circuithub:
Thanks for picking this up! For what it's I've finally finished the Rel8 side of this [here](https://github.com/circuithub/rel8/pull/235).

shane-circuithub:
@tomjaguarpaw, any news on this?

tomjaguarpaw:
It's still on my long list of things to do. If you give me a sense of how important it is to you then that would help me prioritise.

Perhaps the simplest thing to do is just to merge it without tests.

shane-circuithub:
We've been using this in production in CircuitHub for over a month now, the only reason we haven't merged the corresponding Rel8 PR is because it fails CI because the Opaleye that CI tries to build against doesn't have this patch. So it would simplify maintenance for us.

tomjaguarpaw:
OK, I'll aim to merge this at the weekend.  Thanks for prompting me. Please also prompt me in future whenever it would be helpful to you. I can't always get things done quickly but I can always give you an update quickly!

tomjaguarpaw:
Published as https://hackage.haskell.org/package/opaleye-0.9.7.0

shane-circuithub:
Thanks, really appreciate that!

PR #572, opened 2023-10-02
Fixes for invalid SQL generated in WITH queries

shane-circuithub:
This PR contains two fixes pertaining to `WITH` queries. The first wraps an otherwise pointless `SELECT * FROM` around `WITH` queries to workaround PostgreSQL's arbitrary rejection of `WITH foo AS (_) WITH bar AS (_) _` as invalid syntax. The second fixes a more serious issue whereby `withExplicit` is missing a `Rebind` and can generate a "valid" query that silently produces incorrect results.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -25,6 +25,7 @@ import qualified Data.ByteString                  as SBS
 import qualified Data.Text                        as T
 import qualified Data.Time.Compat                 as Time
 import qualified Data.Time.Clock.POSIX.Compat     as Time
+import           Data.Tuple (swap)
 import qualified Database.PostgreSQL.Simple       as PGS
 import qualified Database.PostgreSQL.Simple.Range as R
 import           GHC.Int                          (Int64)
@@ -838,6 +839,12 @@ testNestedWith = it "with nested within with" $ testH with (`shouldBe` expected)
     with = O.with table1Q $ \t -> O.with table2Q $ \u -> (,) <$> t <*> u
     expected = (,) <$> table1data <*> table2data
 
+testWithRebind :: Test
+testWithRebind = it "with (rebinding)" $ testH with (`shouldBe` expected)
+  where
+    with = O.with (fmap swap table1Q) $ \t -> (,) <$> t <*> table2Q
+    expected = (,) <$> fmap swap table1data <*> table2data
+
 -- TODO: This is getting too complicated
 testUpdate :: Test
 testUpdate = it "" $ \conn -> do
@@ -1604,5 +1611,6 @@ main = do
         testWithRecursive
         testWith
         testNestedWith
+        testWithRebind
       describe "relation valued exprs" $ do
         testUnnest
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -832,6 +832,12 @@ testWith = it "with" $ testH with (`shouldBe` expected)
   where with = O.with table1Q $ \t -> (,) <$> t <*> table2Q
         expected = (,) <$> table1data <*> table2data
 
+testNestedWith :: Test
+testNestedWith = it "with nested within with" $ testH with (`shouldBe` expected)
+  where
+    with = O.with table1Q $ \t -> O.with table2Q $ \u -> (,) <$> t <*> u
+    expected = (,) <$> table1data <*> table2data
+
 -- TODO: This is getting too complicated
 testUpdate :: Test
 testUpdate = it "" $ \conn -> do
@@ -1597,5 +1603,6 @@ main = do
       describe "with" $ do
         testWithRecursive
         testWith
+        testNestedWith
       describe "relation valued exprs" $ do
         testUnnest
--- a/src/Opaleye/Internal/Sql.hs
+++ b/src/Opaleye/Internal/Sql.hs
@@ -265,7 +265,12 @@ binary op (select1, select2) = SelectBinary Binary {
   }
 
 with :: PQ.Recursive -> Symbol -> [Symbol] -> Select -> Select -> Select
-with recursive name cols wWith wSelect = SelectWith $ With {..}
+with recursive name cols wWith wSelect =
+  SelectFrom
+    newSelect
+      { attrs = Star
+      , tables = [(NonLateral, SelectWith $ With {..}, Nothing)]
+      }
   where
    wTable = HSql.SqlTable Nothing (sqlSymbol name)
    wRecursive = case recursive of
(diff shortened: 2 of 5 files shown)


tomjaguarpaw:
Thanks! Can you also add tests that fail before this PR and pass after it?

shane-circuithub:
I've added failing tests before the commits that subsequently fix them.

src/Opaleye/Internal/Sql.hs
```diff
@@ -3,7 +3,7 @@
 
 module Opaleye.Internal.Sql where
 
-import           Prelude hiding (product)
+import           Prelude hiding (filter, product)
```
tomjaguarpaw, line 6: Is this change needed?

tomjaguarpaw: Oh, I see, it fixes a warning. That's fine.

shane-circuithub: It's not "needed" here to fix the problems with `WITH`, but it does fix a name shadowing warning that I noticed while working on this. It's a separate commit from the other changes.

Test/Test.hs
```diff
@@ -832,6 +833,18 @@ testWith = it "with" $ testH with (`shouldBe` expected)
...
+testNestedWith = it "with nested within with" $ testH with (`shouldBe` expected)
+  where
+    with = O.with table1Q $ \t -> O.with table2Q $ \u -> (,) <$> t <*> u
+    expected = (,) <$> table1data <*> table2data
+
+testWithRebind :: Test
+testWithRebind = it "with (rebinding)" $ testH with (`shouldBe` expected)
+  where
+    with = O.with (fmap swap table1Q) $ \t -> (,) <$> t <*> table2Q
+    expected = (,) <$> fmap swap table1data <*> table2data
```
tomjaguarpaw, line 846: I was surprised by this failure! For reference, here is the generated code before and after

```sql
SELECT
"cte_renamed0_4" as "result1_6",
"cte_renamed1_4" as "result2_6",
"column10_5" as "result3_6",
"column21_5" as "result4_6"
FROM (SELECT
      *
      FROM (WITH "cte_1" (cte0_2,
                          cte1_2) AS
            (SELECT
             "column1" as "column10_3",
             "column2" as "column21_3"
             FROM "table1" AS "T1")
            SELECT
            *
            FROM (SELECT
                  "cte0_2" as "cte_renamed0_4",
                  "cte1_2" as "cte_renamed1_4"
                  FROM "cte_1" AS "T1") AS "T1",
                 (SELECT
                  "column1" as "column10_5",
                  "column2" as "column21_5"
                  FROM "TABLE2" AS "T1") AS "T2") AS "T1") AS "T1"
```

```sql
SELECT
"cte_renamed0_5" as "result1_7",
"cte_renamed1_5" as "result2_7",
"column10_6" as "result3_7",
"column21_6" as "result4_7"
FROM (SELECT
      *
      FROM (WITH "cte_1" (cte0_2,
                          cte1_2) AS
            (SELECT
             "column21_3" as "rebind0_4",
             "column10_3" as "rebind1_4"
             FROM (SELECT
                   "column1" as "column10_3",
                   "column2" as "column21_3"
                   FROM "table1" AS "T1") AS "T1")
            SELECT
            *
            FROM (SELECT
                  "cte0_2" as "cte_renamed0_5",
                  "cte1_2" as "cte_renamed1_5"
                  FROM "cte_1" AS "T1") AS "T1",
                 (SELECT
                  "column1" as "column10_6",
                  "column2" as "column21_6"
                  FROM "TABLE2" AS "T1") AS "T2") AS "T1") AS "T1"
```

and the difference is specifically

```sql
      FROM (WITH "cte_1" (cte0_2,
                          cte1_2) AS
            (SELECT
             "column1" as "column10_3",
             "column2" as "column21_3"
             FROM "table1" AS "T1")
```

```sql
      FROM (WITH "cte_1" (cte0_2,
                          cte1_2) AS
            (SELECT
             "column21_3" as "rebind0_4",
             "column10_3" as "rebind1_4"
             FROM (SELECT
                   "column1" as "column10_3",
                   "column2" as "column21_3"
                   FROM "table1" AS "T1") AS "T1")
```


Presumably `("column10_3", "column21_3")` is what occurs in the tuple returned from the `Select` but when we swap them that change is not reflected in the generated code. Thus we run a rebind to get them in the correct order (although the rebound names are not used).

We could probably do this without a rebind by using `WITH "cte_1" ("column10_3", "column21_3") AS` but that doesn't matter for this PR. There's probably also a way of architecting Opaleye so that such bugs are harder to write.

tomjaguarpaw:
I should probably add `with` to the property tests. They ought to have caught this.

Thanks @duairc!

tomjaguarpaw:
Released as https://hackage.haskell.org/package/opaleye-0.10.1.1, and deprecated 0.10.1.0.

src/Opaleye/Internal/Sql.hs
tomjaguarpaw: Great, thanks for splitting it out as a separate commit.

Test/Test.hs
shane-circuithub: Even if we could fix the ordering of the `("column10_3", "column21_3")` in the `WITH`, I think we would still want a rebinding. I didn't realise this when I first wrote this code, but `WITH foo (bar, baz) AS (VALUES ('a', 'b', 'c'), ('x', 'y', 'z')) SELECT * FROM foo` generates a query that returns three rows, not two. The parentheses after `foo` only renames the columns, it doesn't restrict their number. This is important for `WITH` because it isn't purely syntactical, in many cases Postgres materializes the expressions therein, so without the rebind (which *does* restrict the number of columns) it can end up doing unnecessary work.

tomjaguarpaw: > generates a query that returns three rows, not two

Do you mean "three *columns*, not two"? Very interesting! In which case, yes, best to be explicit about the rebind. But then maybe we can do without the `ctei_j`?

shane-circuithub: Yeah, I meant to say "columns" there.

I think in theory we can do without the `ctei_j` in that both cases (non-recursive and recursive `WITH`) now end up with a rebinding step (the former because we explicitly make one, and the latter because it goes through `union` or `unionAll` which adds  a rebinding step), but at least `union` doesn't give us a good way to know the names of the symbols it will choose for the rebinding "in advance", which we need for recursive queries because they need to be able to refer to their own columns, the names of which we don't know yet. You could make a version of `union` where the symbols to use in the rebinding are passed in explicitly and similarly for `rebind`, but that was outside the scope of what I was trying to do here.

tomjaguarpaw: Fair enough. It's most important to have something that works and we could revisit the exact form of the generated code later if we want.

PR #573, opened 2023-10-05
Fix bug preventing `orderedAggregate` and `distinctAggregator` from being used together

shane-circuithub:
The first commit here refactors `Aggr` to include the `PrimExpr`s directly, and also gives it a `Traversable` instance allowing them to be replaced with `Symbol`s by `extractAggregateFields`. It also means `Aggr` can be used directly in `AggrExpr` which you had said you'd wanted to do before.

The next commit introduces a failing test demonstrating this bug.

And the final commit fixes this bug, making use of the refactoring done in the first commit. This does add `Eq` and `Ord` constraints to `PrimExpr`, which I think is fine, but if you found that distasteful an alternative would not to rewrite whole `PrimExpr`s in `extractAggregateFields`, but only the `AttrExprs` contained therein. Then you wouldn't need `Eq` and `Ord` for entire `PrimExpr`s, you'd only need to `Eq` and `Ord` for `Symbol` which is more conservative. You can see the comment added to `aggregateU` for more information on the implementation details.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -15,6 +15,7 @@ import qualified Data.Aeson                       as Json
 import qualified Data.Function                    as F
 import           Data.Int (Int32)
 import qualified Data.List                        as L
+import qualified Data.List.NonEmpty               as NE
 import           Data.Monoid                      ((<>))
 import qualified Data.Ord                         as Ord
 import qualified Data.Profunctor                  as P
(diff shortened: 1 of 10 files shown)


tomjaguarpaw:
> This does add Eq and Ord constraints to PrimExpr

I would definitely prefer not to do that, but I haven't looked at the code yet, so I don't want to definitively say what's the best way to solve this problem.

PR #574, opened 2023-10-05
Add support for ordered-set aggregation functions (`WITHIN GROUP`)

shane-circuithub:
This commit adds support for ordered-set aggregation functions such as `mode()` which use the `WITHIN GROUP (ORDER BY _)` syntax. However, it doesn't actually expose any public facing API for this, basically because I don't know yet what that API should be for Opaleye. There are a lot of subtle restrictions on the kinds of arguments you can pass to the ordered-set aggregation functions (e.g., `percentile_disc()` can take either a constant expression or one of the columns from the query we're aggregating over, but only if it's part of the `GROUP BY` clause) that would be hard to capture in the types that Opaleye currently uses.
    
What this commit does is add just enough internals for me to able to experiment with a limited API in Rel8 for ordered-set aggregation functions that might be subject to change in the future. But I don't want to add something half-baked to Opaleye's public facing API just yet.

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -15,6 +15,7 @@ import qualified Data.Aeson                       as Json
 import qualified Data.Function                    as F
 import           Data.Int (Int32)
 import qualified Data.List                        as L
+import qualified Data.List.NonEmpty               as NE
 import           Data.Monoid                      ((<>))
 import qualified Data.Ord                         as Ord
 import qualified Data.Profunctor                  as P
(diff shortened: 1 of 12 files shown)


shane-circuithub:
(Note that this work is based on top of #573.)

tomjaguarpaw:
Is it possible to do this without https://github.com/tomjaguarpaw/haskell-opaleye/pull/573? I'm happy to expose these internals for you to play around with but I won't be able to review https://github.com/tomjaguarpaw/haskell-opaleye/pull/573 quickly.

PR #575, opened 2023-10-05
Refactor `Aggr` to include the expressions it aggregates

shane-circuithub:
This PR makes several (hopefully uncontroversial) changes to Opaleye's internal data types for aggregation that are needed for followup work.

The first change basically changes:

```haskell
newtype Aggregator a b =
  Aggregator (PM.PackMap (HPQ.Aggr, HPQ.PrimExpr) HPQ.PrimExpr a b)

data PrimQuery' a = ...
                  | Aggregate (Bindings (HPQ.Aggr, HPQ.Symbol))
                              (PrimQuery' a)

data PrimExpr   = ...
                | AggrExpr  AggrDistinct AggrOp PrimExpr [OrderExpr] (Maybe PrimExpr)

data Aggr
  = GroupBy
  | Aggr
      { aggrOp :: !AggrOp
      , aggrOrder :: ![OrderExpr]
      , aggrDistinct :: !AggrDistinct
      , aggrFilter :: !(Maybe PrimExpr)
      }
  deriving (Show, Read)
```

to:

```haskell
newtype Aggregator a b =
  Aggregator (PM.PackMap HPQ.Aggregate HPQ.PrimExpr a b)

data PrimQuery' a = ...
                  | Aggregate (Bindings (HPQ.Aggregate' HPQ.Symbol))
                              (PrimQuery' a)

data PrimExpr   = ...
                | AggrExpr  Aggr

type Aggregate = Aggregate' PrimExpr

data Aggregate' a = GroupBy a | Aggregate (Aggr' a)
  deriving (Functor, Foldable, Traversable, Show, Read)

type Aggr = Aggr' PrimExpr

data Aggr' a = Aggr
  { aggrOp :: !AggrOp
  , aggrExprs :: ![a]
  , aggrOrder :: ![OrderExpr]
  , aggrDistinct :: !AggrDistinct
  , aggrFilter :: !(Maybe PrimExpr)
  }
  deriving (Functor, Foldable, Traversable, Show, Read)
```

[...]

PR #576, opened 2023-10-05
Add support for ordered-set aggregation functions (`WITHIN GROUP`)

shane-circuithub:
This commit adds support for ordered-set aggregation functions such as `mode()` which use the `WITHIN GROUP (ORDER BY _)` syntax. However, it doesn't actually expose any public facing API for this, basically because I don't know yet what that API should be for Opaleye. There are a lot of subtle restrictions on the kinds of arguments you can pass to the ordered-set aggregation functions (e.g., `percentile_disc()` can take either a constant expression or one of the columns from the query we're aggregating over, but only if it's part of the `GROUP BY` clause) that would be hard to capture in the types that Opaleye currently uses.

What this commit does is add just enough internals for me to able to experiment with a limited API in Rel8 for ordered-set aggregation functions that might be subject to change in the future. But I don't want to add something half-baked to Opaleye's public facing API just yet.

### DIFF
--- a/src/Opaleye/Aggregate.hs
+++ b/src/Opaleye/Aggregate.hs
@@ -38,6 +38,7 @@ module Opaleye.Aggregate
 import           Control.Applicative (pure)
 import           Data.Profunctor     (lmap)
 import qualified Data.Profunctor as P
+import           Data.Profunctor.Product.Default (def)
 
 import qualified Opaleye.Internal.Aggregate as A
 import           Opaleye.Internal.Aggregate (Aggregator, orderAggregate)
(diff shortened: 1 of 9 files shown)


shane-circuithub:
This is based on the refactoring in #575, but no longer depends on the fixes proposed in #577 or #578.

tomjaguarpaw:
Yup, all good. Likewise split up a bit and merged separately.

----

Due to the splitting tool all the commits get attributed to me. Sorry about that. I try to credit you on all the commits, e.g. https://github.com/tomjaguarpaw/haskell-opaleye/commit/1367e701e665e1d15395d3cf30ce3aba5ee87333. If you feel strongly about this please let me know and I'll see what I can do to fix  the tool.

shane-circuithub:
No worries at all, I don't care about the attribution, just happy to have this merged so I can play around with it! Thanks for being so responsive in the last week with all the PRs.

tomjaguarpaw:
You're welcome. Thanks for the great PRs!

PR #577, opened 2023-10-05
Fix bug preventing `orderedAggregate` and `distinctAggregator` from being used together (one approach)

shane-circuithub:
This PR contains two commits.

The first commit adds a (failing) test case demonstrating the inability to combine `orderedAggregate` and `distinctAggregator` together, even in the limited cases allowed by PostgreSQL. The problem is that PostgreSQL only allows this if the expressions given in the `ORDER BY` clause match the expressions given as arguments to the aggregation function.

The second commit fixes this bug. It does so by adding `Eq` and `Ord` constraints to `PrimExpr` and using them to detect identical expressions anywhere within the `Aggregate`, and when we rename the expressions we maintain a `Map` of already-renamed expressions such that identical expressions will be renamed to the same symbol, fulfilling PostgreSQL's restriction.

### DIFF
--- a/opaleye.cabal
+++ b/opaleye.cabal
@@ -29,8 +29,9 @@ library
       aeson               >= 0.6     && < 2.3
     , base                >= 4.9     && < 4.19
     , base16-bytestring   >= 0.1.1.6 && < 1.1
-    , case-insensitive    >= 1.2     && < 1.3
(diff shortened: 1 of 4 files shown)


shane-circuithub:
This is based on the refactoring in #575.

shane-circuithub:
Closing in favour of #578.

PR #581, opened 2023-10-11
Relax the type of the Unpackspec for makeAggrExplicit

shane-circuithub:


src/Opaleye/Aggregate.hs
```diff
@@ -195,7 +195,9 @@ jsonAgg = A.makeAggr HPQ.JsonArr
 stringAgg :: F.Field T.SqlText
           -> Aggregator (F.Field T.SqlText) (F.Field T.SqlText)
 stringAgg delimiter =
-  lmap (\a -> (a, delimiter)) $ A.makeAggrExplicit def HPQ.AggrStringAggr
+  A.makeAggrExplicit
+    (U.unpackspecField <* lmap (const delimiter) U.unpackspecField)
+    HPQ.AggrStringAggr
```
tomjaguarpaw, line 200: This is just a refactoring, right? I think I actually prefer the other way, since shouldn't we morally be able to have

```.hs
stringAgg' :: Aggregator (F.Field T.SqlText) F.Field T.SqlText) (F.Field T.SqlText)
stringAgg' = A.makeAggrExplicit def HPQ.AggrStringAgg
```

and then

```haskell
stringAgg :: F.Field T.SqlText
          -> Aggregator (F.Field T.SqlText) (F.Field T.SqlText)
stringAgg delimiter = lmap (\a -> (a, delimiter)) stringAdd'
```

I don't _think_ there's anything stopping the delimiter being dynamic (i.e. a value depending on the execution of the query -- although I haven't checked that).

shane-circuithub: The change to `stringAgg` is just a refactoring, yes. And you're right that the delimiter can be dynamic, I just didn't want to break the API of the existing `stringAgg`.

But I do need to relax the type of the `Unpackspec` for `makeAggrExplicit` (which I'm using in Rel8 for ordered set aggregation functions). Maybe `Unpackspec` is the wrong thing to use here, but basically I'm using it to select the `PrimExpr`s that will be passed as the argument list to the aggregation function. I'm open to other suggestions

tomjaguarpaw: OK, let's merge this to unblock you and we can attend to other considerations later if we want.

PR #582, opened 2023-10-13
Delete no longer relevant tutorial sections

jhrcek:
All of these "would you like to volunteer to implement this missing feature" comments seem no longer relevant, because the mentioned features have already been implemented (please correct me if I'm wrong).

### DIFF
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -644,9 +644,7 @@ String' into a 'Field Int64'.
 Outer join
 ==========
 
-Opaleye supports left joins.  (Full outer joins and right joins are
-left to be added as a simple starter project for a new Opaleye
-contributor!)
+Opaleye supports left joins.
 
 Because left joins can change non-nullable fields into nullable
 fields we have to make sure the type of the output supports
--- a/Doc/Tutorial/TutorialBasic.lhs
+++ b/Doc/Tutorial/TutorialBasic.lhs
@@ -171,8 +171,7 @@ example we might want to discard the "address" field of our
 Projection gives us our first example of using "do notation" to
 write Opaleye queries.
 
-Here we run the `personSelect` passing in () to signify "zero
-arguments".  We pattern match on the results and return only the
+Here we run the `personSelect`, pattern match on the results and return only the
 fields we are interested in.
 
 > nameAge :: Select (Field SqlText, Field SqlInt4)
--- a/Doc/Tutorial/TutorialBasicMonomorphic.lhs
+++ b/Doc/Tutorial/TutorialBasicMonomorphic.lhs
@@ -67,12 +67,6 @@ manipulation tutorial you can see an example of when they might differ.
 >                                       , tableField "age"
 >                                       , tableField "address" ))
 
-> personTable' :: Table (Field SqlText, Field SqlInt4, Field SqlText)
->                       (Field SqlText, Field SqlInt4, Field SqlText)
-> personTable' = table "personTable" (p3 ( tableField "name"
->                                        , tableField "age"
->                                        , tableField "address" ))
-
 By default, the table `"personTable"` is looked up in PostgreSQL's
 default `"public"` schema. If we wanted to specify a different schema we
 could have used the `tableWithSchema` constructor instead of `table`.
@@ -148,9 +142,6 @@ them.
 >
 > instance Default Unpackspec BirthdayField BirthdayField where
 >   def = birthdayFieldDef
-
-Naturally this is all derivable using `Generic` or Template Haskell,
-but no one's bothered to implement that yet.  Would you like to?
 
 Then we can use 'table' to make a table on our record type in exactly
 the same way as before.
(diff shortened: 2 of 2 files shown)


Doc/Tutorial/TutorialBasicMonomorphic.lhs
```diff
@@ -67,12 +67,6 @@ manipulation tutorial you can see an example of when they might differ.
 >                                       , tableField "age"
 >                                       , tableField "address" ))
 
-> personTable' :: Table (Field SqlText, Field SqlInt4, Field SqlText)
```
jhrcek, line 70: This definition is not used anywhere, so just deleting it.

Doc/Tutorial/TutorialBasic.lhs
```diff
@@ -171,8 +171,7 @@ example we might want to discard the "address" field of our
 Projection gives us our first example of using "do notation" to
 write Opaleye queries.
 
-Here we run the `personSelect` passing in () to signify "zero
```
jhrcek, line 174: "passing ()" seems like leftover from Arrow style api?

Doc/Tutorial/TutorialBasicMonomorphic.lhs
```diff
@@ -149,9 +143,6 @@ them.
 > instance Default Unpackspec BirthdayField BirthdayField where
 >   def = birthdayFieldDef
 
-Naturally this is all derivable using `Generic` or Template Haskell,
-but no one's bothered to implement that yet.  Would you like to?
-
```
tomjaguarpaw, line 154: This comment is specifically about deriving for *monomorphic* record types, so the comment is still true.

Doc/Tutorial/TutorialBasicMonomorphic.lhs
```diff
@@ -297,9 +288,6 @@ purpose, which is just a notational convenience.
 >   def = BirthdayFieldNullable <$> P.lmap bdNameField D.def
 >                                <*> P.lmap bdDayField  D.def
 
-Again, this is all derivable using `Generic` or Template Haskell, if
-someone would take the time to implement it.
-
```
tomjaguarpaw, line 302: This comment is specifically about deriving for *monomorphic* record types, so the comment is still true.

Doc/Tutorial/TutorialBasic.lhs
tomjaguarpaw: Yes, thanks.

Doc/Tutorial/TutorialBasicMonomorphic.lhs
tomjaguarpaw: I suspect it was supposed to be an example that used `tableWithSchema`. If you'd like to change it to that then that would be helpful, otherwise I'll do it myself when I get round to it.

tomjaguarpaw (commented):
Thanks very much for this. A couple of the "not implemented yet" comments are still true, actually. Could you please reinstate them?

Doc/Tutorial/TutorialBasicMonomorphic.lhs
jhrcek: Sure, I changed it to tableWithSchema.

Doc/Tutorial/TutorialBasicMonomorphic.lhs
jhrcek: Ok, change reverted.

Doc/Tutorial/TutorialBasicMonomorphic.lhs
jhrcek: Reverted.

jhrcek:
Hm, didn't realize the generic stuff is not implemented for monomorphic types.
Could you please sketch what exactly should be implemented using generics?
I mean in the form like 
```
instance (... some context) => Default what1 what2 what3 where ...
```
I might give it a try as a learning project :smile: 
I understand the general technique of using default instance methods + DeriveAnyClass, just not sure what exactly you mean by "this is all derivable by Generics"..

tomjaguarpaw:
Thank you!

tomjaguarpaw:
> Hm, didn't realize the generic stuff is not implemented for monomorphic types.
> Could you please sketch what exactly should be implemented using generics?

Could you clarify the question? What needs to be implemented is derivation of exactly the code from that section of the tutorial, that is, define a "birthday" type with two fields, a "text" and a "date" and generate from it the following code

```.lhs
> data BirthdayField = BirthdayField { bdNameField :: Field SqlText
>                                      , bdDayField  :: Field SqlDate }
>
> data Birthday = Birthday { bdName :: String, bdDay :: Day }
>
> birthdayFieldDef ::
>   (Applicative (p BirthdayField),
>    P.Profunctor p,
>    Default p (Field SqlText) (Field SqlText),
>    Default p (Field SqlDate) (Field SqlDate)) =>
>   p BirthdayField BirthdayField
> birthdayFieldDef = BirthdayField <$> P.lmap bdNameField D.def
>                                    <*> P.lmap bdDayField  D.def
>
> instance Default Unpackspec BirthdayField BirthdayField where
>   def = birthdayFieldDef
```

PR #583, opened 2023-10-15
Add support for `WITH _ AS MATERIALIZED (_)` queries

shane-circuithub:
Before PostgreSQL 12, `WITH` was always materialized, but since PostgreSQL 12 this is not necessarily the case (it's up to the query planner).

However, a new syntax was added for explicitly specifying whether a `WITH` query should be `MATERIALIZED` or `NOT MATERIALIZED`. This commit adds support for this to Opaleye's core, and adds the function `withMaterialized` to the user-facing API.


tomjaguarpaw:
Released in https://hackage.haskell.org/package/opaleye-0.10.2.0

PR #586, opened 2024-01-13
Avoid rebinding PrimExprs that were created in the Aggregator itself

tomjaguarpaw:


tomjaguarpaw:
I'd like to understand why @duairc made the change https://github.com/circuithub/rel8/pull/305#discussion_r1460771982, because I'm concerned that this PR has an impact I'm not aware of. Besides that, this PR is ready to go.

tomjaguarpaw:
Still hoping I'll hear back from @duairc about this before merge.

shane-circuithub:
Hey @tomjaguarpaw, sorry I took so long to get back to you. Yes, the changes you see on the Rel8 side are all just to account for the additional `Table Expr` constraint we now incur. `countRows` doesn't really care about the specific layout of columns though which is why we just map `const true` over the input query.

tomjaguarpaw:
Thanks, that's OK. Released at https://hackage.haskell.org/package/opaleye-0.10.2.1

PR #588, opened 2024-04-08
IsSqlType instance is needed if the field is an array of enum

stevemao:
when toFields is used on sql array, it uses this instance

```
(Default ToFields a (Field_ n b), IsSqlType b) => Default ToFields [a] (Field (SqlArray_ n b))
```

Which cast the value to the enum type


tomjaguarpaw:
Thanks! At some point I'll fold this into `EnumMapper`.

tomjaguarpaw:
I incorporated this into `EnumMapper` in https://github.com/tomjaguarpaw/haskell-opaleye/commit/91727ec6af166e6a5c9bdfa8cced4f41bcf9dc93

PR #589, opened 2024-04-08
add showSqlTypeWithSchema for types in a schema

stevemao:
- similar to enumMapperWithSchema
- enums needs an instance of IsSqlType if used with an array
- enums can have a schema


stevemao:
I can, but other APIs such as `enumMapperWithSchema` has a spot for schema. This is to make it consistent with these APIs.

tomjaguarpaw:
I see. Then in that case I think `showSqlTypeWithSchema` wants just to be a function

```.hs
showSqlTypeWithSchema :: (String, String) -> String
showSqlTypeWithSchema (schema, type_) =
  render (doubleQuotes (text schema) <> text "." <> doubleQuotes (text type_))
```

that you use like

```
instance IsSqlType Whatever where
  showSqlType = showSqlTypeWithSchema ("schema", "type_name")
```

Making it a class method means that *all* instances have to implement it, which is not what we want because not all type names have a schema.

stevemao:
Happy with a function to make it simple. But I have some questions

`showSqlTypeWithSchema` seems to be quite generic. Maybe we can call it `sqlTypeWithSchema`?

> Making it a class method means that all instances have to implement it, which is not what we want because not all type names have a schema.

Not really? the instance can choose to implement either method `{-# MINIMAL showSqlType | showSqlTypeWithSchema #-}`

tomjaguarpaw:
> >  Making it a class method means that all instances have to implement it, which is not what we want because not all type names have a schema.
>
> Not really? the instance can choose to implement either method {-# MINIMAL showSqlType | showSqlTypeWithSchema #-}

But what if an instance choose not to implement it but then a user calls it on that instance? Run time crash, presumably.

stevemao:
Ok, definitely not ideal even though it's an internal module

tomjaguarpaw:
Thanks, I tweaked this and merged it in https://github.com/tomjaguarpaw/haskell-opaleye/commit/f012b275608e5109f913be61260cdbdf6149d088

tomjaguarpaw:
Released as https://hackage.haskell.org/package/opaleye-0.10.3.0

PR #615, opened 2025-01-29
Allow limits and offsets by dynamic expressions

shane-circuithub:

PR #616, opened 2025-01-29
Allow limits and offets to work with Fields not just Ints

tomjaguarpaw:
This is https://github.com/tomjaguarpaw/haskell-opaleye/pull/615 but with minor adjustments (in particular, we don't already call anything "dynamic" in Opaleye, so I changed the names to `limitField` and `offsetField`).

All credit due to @shane-circuithub


tomjaguarpaw:
Thanks @shane-circuithub!

PR #620, opened 2025-07-10
Instances for `Int16`

### LINKED ISSUE #505: No Default ToFields Int16?
I am trying to write an application atop an existing PostgreSQL database that hast `smallint` values. To my understanding, they should be mapped onto Haskell's `Int16` type. However, I get a typeclass error `No instance for (OE.DefaultFromField OE.SqlInt2 Int16)`. A look into `Opaleye.ToFields` shows that there indeed is no instance for `Int16`.  Is this intentional for some reason, or could I simply add an instance to make it work?

fpringle:
Adds missing instances (`Default ToFields`, `DefaultFromField` and `SqlNum`) for `Int16`/`SqlInt2`.
Closes #505.

Mostly mechanical copying of the instances for `Int32` etc. The only un-obvious part is the type of `sumInt2` - the PostgreSQL [docs](https://www.postgresql.org/docs/17/functions-aggregate.html#:~:text=sum%20(%20smallint%20)%20%E2%86%92%20bigint) say that `sum(smallint)` has type `bigint`.


tomjaguarpaw:
Thanks! Do you need a Hackage release too?

tomjaguarpaw:
OK, here you go! https://hackage.haskell.org/package/opaleye-0.10.6.0

PR #622, opened 2025-07-14
Add a typed version of `null` so that Postgres doesn't freak out on some `cast`s

### LINKED ISSUE #621: Weird behaviour in nullable columns
Hi, I'm noticing a weird behaviour, which seems to be more of a quirk of PostgreSQL than anything else, but I thought I'd record it here.

Say we have a (slightly contrived) `Field`:

```haskell
ghci> field :: FieldNullable SqlJson = ifThenElse (toFields True) null null
ghci> Column prim = field
ghci> ppSqlExpr $ sqlExpr defaultSqlGenerator $ prim
CASE WHEN CAST(TRUE AS boolean) THEN NULL ELSE NULL END
```

Curiously, when we try to insert this value in SQL, we get a type error:

```sql
CREATE TEMP TABLE test_table (json_nullable json null);

-- this is fine
=> insert into test_table values (null);
INSERT 0 1
=> insert into test_table values (CASE WHEN CAST(TRUE AS boolean) THEN NULL ELSE NULL END);
ERROR:  column "json_nullable" is of type json but expression is of type text
LINE 1: insert into test_table values (CASE WHEN CAST(TRUE AS boolea...
                                       ^
HINT:  You will need to rewrite or cast the expression.
```

I realise this is an issue with SQL, not Opaleye, but is there any way to mitigate this ~in the `ToFields` instance for `SqlJson`~?

---

Edit: I re-tried the SQL code above with `int` instead of `json`, same issue.

fpringle:
Closes #621, see that issue for detailed discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+* Added `unsafeCastSqlType`, `typedNull` and `untypedNull`, allowing
+  workaround for
+  [#621](https://github.com/tomjaguarpaw/haskell-opaleye/issues/621).
+  Thanks to @fpringle.
+
 ## 0.10.6.0
 
 * Added `int2` functions and instances (thanks to @fpringle)
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1527,6 +1527,17 @@ testSetAggregate = do
         expectation :: Double
         expectation = 2.5
 
+-- See https://github.com/tomjaguarpaw/haskell-opaleye/issues/621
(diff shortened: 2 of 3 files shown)


tomjaguarpaw:
Thanks, I edited it a bit.

tomjaguarpaw:
Released in https://hackage.haskell.org/package/opaleye-0.10.7.0

fpringle:
Thanks @tomjaguarpaw!

PR #625, opened 2025-08-14
Add `Eq` and `Ord` instances for `TableIdentifier`

fpringle:
Tiny PR adding an `Ord` instance to `TableIdentifier` so that I can use it as a key in a `Map` in [effectful-opaleye](https://github.com/fpringle/effectful-postgresql/tree/main/effectful-opaleye).

I understand there's an argument to be made against the `Eq`, since there's a context in which `schema.table` and `table` might be considered equal, despite not comparing equal as Haskell values. If this is an issue, I'd be happy to swap the instances out for a `Hashable` instance (the `hashable` package is already a transitive dependency, so it wouldn't add to the footprint).

### DIFF
--- a/src/Opaleye/Internal/PrimQuery.hs
+++ b/src/Opaleye/Internal/PrimQuery.hs
@@ -29,7 +29,7 @@ data SemijoinType = Semi | Anti deriving Show
 data TableIdentifier = TableIdentifier
   { tiSchemaName :: Maybe String
   , tiTableName  :: String
-  } deriving Show
+  } deriving (Show, Eq, Ord)
 
 tiToSqlTable :: TableIdentifier -> HSql.SqlTable
 tiToSqlTable ti = HSql.SqlTable { HSql.sqlTableSchemaName = tiSchemaName ti


tomjaguarpaw:
Hmm, I'm a bit dubious about using this internal datatype in another library. Is there some public data type you could use instead?

fpringle:
Not in Opaleye, but now you mention it I do notice that postgresql-simple has [QualifiedIdentifier](https://hackage.haskell.org/package/postgresql-simple-0.6.5/docs/Database-PostgreSQL-Simple-Types.html#t:QualifiedIdentifier) which seems to be isomorphic. I'll still need to depend on `TableIdentifier`, which I get from `Insert`s, `Delete`s and `Update`s in order to keep a running tally of how many DB operations are being dispatched.

tomjaguarpaw:
I see. [`Table`](https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/src/Opaleye/Internal/Table.hs#L47-L49) could actually incorporate a type like that. It doesn't have to be factored the way that it currently is.

For now can you just use your own custom type as the `Map` key and marshall between it and `TableIdentifier` when necessary? If you use that for six months and you're sure it's the right type for the job come back and I'll think again about what we can expose from the public API to do the job.

fpringle:
I'd imagine you could replce `TableIdentifier` with `QualifiedIdentifier` pretty easily, since they're identical (except field names), and internal.
I'll do as you suggested, except I'll just use `QualifiedIdentifier` as the map key.
No matter what type I use, I'll need to be able to get some kind of table identifier from the `Insert`, `Delete` and `Update` types.

fpringle:
In case you're curious, [this](https://github.com/fpringle/effectful-postgresql/pull/4/files#diff-3fb23085b5e5679b723826a6454b5cf48c93fa7ab02c01fba6bb19140c595fb8) PR contains the code that uses `QualifiedIdentifier` as a map key to count DB operations.

PR #626, opened 2026-02-27
Add AnyExpr for ANY expressions

simmsb:
Heya, I've come across a need for `<expr> = ANY(...)` support due to the postgres version I'm using not being smart enough to use an index when the [`array_position` implementation](https://github.com/tomjaguarpaw/haskell-opaleye/blob/74c1f9db2a152b077f39668b265c015a48888bb4/src/Opaleye/Operators.hs#L445) of `sqlElem` is used.

I'm not sure if this is the correct solution to implementing it (but I think it is as `<op> ANY(...)` seems to be a special form in (postgres?)sql too)

I've been using this patch privately for a month or so now.

### DIFF
--- a/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
+++ b/src/Opaleye/Internal/HaskellDB/Sql/Default.hs
@@ -124,6 +124,10 @@ defaultSqlExpr gen expr =
                 (leftE, paren rightE)
               _ -> (paren leftE, paren rightE)
         in BinSqlExpr (showBinOp op) expL expR
+      AnyExpr op e1 e2 ->
+        let leftE = sqlExpr gen e1
+            rightE = sqlExpr gen e2
+        in AnySqlExpr (showBinOp op) (ParensSqlExpr leftE) rightE 
       UnExpr op e      -> let (op',t) = sqlUnOp op
                               e' = sqlExpr gen e
                            in case t of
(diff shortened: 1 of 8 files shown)


tomjaguarpaw:
Thanks! This looks good. Probably needs function exposed for it in `Opaleye.Operators` and a test?

tomjaguarpaw:
The test should go in here: https://github.com/tomjaguarpaw/haskell-opaleye/blob/master/Test/Test.hs

simmsb:
I've updated `sqlElem` to just use `AnyExpr`, there's a test for `sqlElem` already and that continues to pass when I run the tests locally.

I think your CI is upset about the copyright year not being 2026, I've updated that so that CI can progress.

tomjaguarpaw:
> I've updated `sqlElem` to just use `AnyExpr`

Thanks, but that's a change in behaviour so would require a major version bump. Let's make it a new function (`sqlElemAny`?) so that it's opt-in and doesn't change the code of anyone who's already using this (and add a test for that new function).

tomjaguarpaw:
Thank you! I'll get to this in due course.

src/Opaleye/Internal/HaskellDB/Sql/Default.hs
```diff
@@ -124,6 +124,10 @@ defaultSqlExpr gen expr =
                 (leftE, paren rightE)
               _ -> (paren leftE, paren rightE)
         in BinSqlExpr (showBinOp op) expL expR
+      AnyExpr op e1 e2 ->
+        let leftE = sqlExpr gen e1
+            rightE = sqlExpr gen e2
+        in AnySqlExpr (showBinOp op) (ParensSqlExpr leftE) rightE
```
tomjaguarpaw:
```suggestion
        in AnySqlExpr (showBinOp op) (ParensSqlExpr leftE) rightE
```

tomjaguarpaw:
Thanks, released as: https://hackage.haskell.org/package/opaleye-0.10.8.0

PR #629, opened 2026-05-29
Add ON CONFLICT DO UPDATE SET support (doUpdate, doUpdateEasy, doUpdateAll)

alex-tan:
## Summary

- Extends `OnConflict` with a `DoUpdate` variant carrying a conflict target and a list of `(column, expr)` assignments
- Adds `ExcludedAttrExpr` to `PrimExpr` so the `EXCLUDED` pseudo-table renders correctly in SQL
- Exposes three smart constructors in `Opaleye.Manipulation`:
  - `doUpdate` — update function in write-type space; optional columns need explicit `Just`
  - `doUpdateEasy` — update function in read-type space; no `Just` wrapping needed
  - `doUpdateAll` — replaces every column with the corresponding `EXCLUDED` value (most common upsert pattern)
- Adds tests for all three variants

### DIFF
--- a/Test/Test.hs
+++ b/Test/Test.hs
@@ -1045,10 +1045,10 @@ testInsertConflict = it "inserts with conflicts" $ \conn -> do
 
         runSelectTable10 conn = O.runSelect conn (O.selectTable table10)
 
--- Helper shared by the three doUpdate tests: inserts initial rows,
--- runs an upsert with the given conflict action, then checks the result.
-runUpsertTest :: O.OnConflict -> PGS.Connection -> IO ()
-runUpsertTest onConflict conn = do
+-- Helper shared by the doUpdate tests: inserts initial rows, runs an
+-- upsert with the given conflict action, then checks the result.
+runUpsertTest :: O.OnConflict -> [UpsertRow' Int Int] -> PGS.Connection -> IO ()
+runUpsertTest onConflict expected conn = do
   _ <- O.runDelete_ conn O.Delete { O.dTable     = upsertTable
                                   , O.dWhere     = const (O.toFields True)
                                   , O.dReturning = O.rCount }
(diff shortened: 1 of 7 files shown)


tomjaguarpaw:
Cool! I'll have a look it due course. Feel free to ping if you think I may have forgotten.

alex-tan:
Thanks!

tomjaguarpaw:
FYI I haven't forgotten about this, but I have a queue of other things to get to first.

tomjaguarpaw:
I have a bit of time to look at this. I'm curious about this choice of API:

```.hs
doUpdate :: ( D.Default U.Unpackspec conflictCols conflictCols
            , D.Default U.Unpackspec fieldsR fieldsR
            )
         => T.Table fieldsW fieldsR
         -> (fieldsR -> conflictCols)
         -- ^ Columns forming the conflict target
         -> (fieldsR -> fieldsW)
         -- ^ Update function; receives the @EXCLUDED@ pseudo-row
         -> HSql.OnConflict
```

Don't you also want a way to refer to the existing row? So perhaps the update function should be `fieldsR -> fieldsR -> fieldsW`?

tomjaguarpaw:
If in principle one would want that but you personally don't want that then I'm happy to incorporate your change, but maybe with the name `doUpdateWithExcluded`?

alex-tan:
Sorry for the delayed response and thanks for the review.

That makes sense. I just pushed two commits addressing your comment and a few other things.

Give doUpdate access to the existing row

- Changed doUpdate/doUpdateEasy to take fieldsR -> fieldsR -> fieldsW
- doUpdateAll becomes doUpdateEasy table t (\_ excluded -> excluded)
- Fixes an issue where there was no way to accumulate onto the existing value or to leave a column unchanged. Now Nothing on an optional field means DEFAULT, not "leave alone".
- Renamed ExcludedAttrExpr Attribute → QualifiedAttrExpr String Attribute, since both rows must be qualified on the RHS of DO UPDATE SET — the target table and excluded are both in scope, so a bare column name is ambiguous. Existing row is qualified by the bare table name.

Allow expressions as the conflict target

- SqlConflictColumns now holds SqlExprs instead of SqlColumns, so ON CONFLICT ((lower(sku))) can work against expression indexes.
- Target is now NonEmpty, since DO UPDATE with no target is a syntax error.
- Removed SqlConflictConstraint which was unreachable (maybe something that could be implemented later if necessary?).

